PR #24, opened 2018-05-21
sfgovpl theme

isramv:
This PR provides a new sfgov theme with pattern lab.


jszwedko:
Apologies, this is somewhat of a drive-by comment as I have limited familiarity with Pattern Lab, but I was somewhat surprised to see the entirety of this embedded inside of a Drupal theme rather than as a standalone repository that generated versioned assets which a Drupal theme might consume.

I anticipate the benefits of this as a separation of concerns:

* designers and others could iterate on the pattern outside of the context of the sf.gov website
* a separate build, test, deploy process
* easier to use the patterns across web applications (e.g. businessportal.sfgov.org, housing.sfgov.org, etc.)

Am I completely off the mark?

Jaesin:
Hi Jesse, 
Thanks for looking it over. You've got it right. The pattern lab will be in it's own repo. We'll ask Josh to set up the repository.

We're still sorting out how to consume the pattern library in the drupal theme. It might just be a matter of getting SF Gov set up as an org on npm.

isramv:
Hi Jesse,

We discussed all this with Josh on our weekly checkin today, but to catch you up....

The idea behind this PR is to generate the structure and build the custom components on top of the uswds.css and uswds.js and at somepoint, when the Pattern Library is robust enough remove the pattern-lab folder and add it to the project as depencency of the theme.

The USWDS uses Fractal, instead of Pattern Lab. Fractal is pretty much abandoned (the latest PR is from last year: https://github.com/frctl/twig/pull/7). I tried to refactor the .njk templates to twig using the following frctl plugin (https://github.com/frctl/twig). Unfortunately, it's not supporting twig macros which are by the way, heavily used by the menus. So due to its incompatibility and lack of support, we are borrowing the styles and applying them to Pattern Lab, which is much more active as a project.

The USWDS implementation is also pretty shallow– there are just not too much components available: https://designsystem.digital.gov/components/

That's everything, which isn't super scalable for you guys long term.

We still need to set up and configure a separate `build, test, deploy process` which is not my field of work, and I'll be working on this with Jaesin, who y ou met when you came into the Chapter Three office.

My work can and will be extrapolated at a later time by him who is more comfortable in this area.

This theme has been designed so, the twig, scss, and js files are placed in the pattern-lab folder but nothing Drupal related. so a separation in the future would be fairly easy.

So in summary, we are going to do it! Just not right this second, as our intent at this time is not to have people using the current design. We are still working on it though, because we want it to be ready once the branding effort with DesignMap is more solid!

jszwedko:
Thanks for the additional context @isramv !

I still worry a bit about the Pattern Lab implementation becoming overly coupled with the sf.gov Drupal implementation and being more difficult to extract later. I also foresee it rather complicating the build process in this project and possibly making it more difficult to iterate on the Pattern Lab assets with DesignMap or other partners. I'll try not to belabor the point since it's already been discussed, but do we see separating it early as overly burdensome? We could hypothetically just link to "latest" Pattern Lab assets in the Drupal theme.

isramv:
Hey @jszwedko,

I'm already working on the separation of the pattern-lab from the theme.
later today I'll send a new PR.

Thanks.

jszwedko:
👍 I also talked with Josh out-of-band who filled me in. Thanks!

PR #29, opened 2018-05-22
SG-54 Roles

arlina-espinoza:
Adding drupal and group roles and configuration.


web/modules/custom/sfgov_departments/src/SFgovDepartment.php
```diff
@@ -3,6 +3,7 @@
 namespace Drupal\sfgov_departments;
```
josh-chou: Could you add simple unit tests to cover the SFgovDepartment functionality?

arlina-espinoza:
Hi @josh-chou, I've added a test class that checks the functionality of the SFgovDepartment class. Could you review? Thanks!

PR #32, opened 2018-05-24
[SG-101] Remove text field from Document paragraph.

jacine:
This field isn't supposed to be here. Should only be on the printable/downloadable form, which remains.


jacine:
And.... I removed it from the wrong Paragraph accidentally. Closing this and will open a corrected PR.

PR #37, opened 2018-05-24
[SG-26] - sitename and logo.

isramv:
First integration between the pattern library and the branding block.


isramv:
@josh-chou I am going to send a new PR, I thought those configurations wouldn't affect the config import.
plus those configurations are not necessary right now.

PR #40, opened 2018-05-29
[SG-115] Install Metatag Verification with Google config.

jacine:


josh-chou:
Jacine, can you shed light on my question about the metatag naming?  Thanks!

jacine:
Sure @josh-chou!  Where did you ask the question?

josh-chou:
Hmm... I guess my question is only viewable to me.  That might be why it's marked pending.
Let me try a different way of commenting....

config/metatag.metatag_defaults.global.yml
```diff
@@ -7,5 +7,6 @@ _core:
 id: global
 label: Global
 tags:
-  canonical_url: '[current-page:url]'
   title: '[current-page:title] | [site:name]'
+  canonical_url: '[current-page:url]'
+  google: 32Y54vvXJK0dXVMHal_Axqvcn7FLe1ymoM1x478IpV4
```
josh-chou: Is this supposed be the name of the meta tag? If so, then it should be "google-site-verification"?

jacine: No, it's actually Metatag module's internal configuration key.  It gets mapped to `google-site-verification` when processed.  The relevant code is [here](https://cgit.drupalcode.org/metatag/tree/metatag_verification/metatag_verification.metatag.inc#n42) if you're curious.

josh-chou: Wonderful!  Thanks for the explanation.

PR #42, opened 2018-05-29
[SG-109] Fix “Ways to get Help” field and Paragraph.

jacine:


jacine:
I've manually tested this locally and the PR seems fine.  Looks like it's probably failing due to some Pantheon & CI issues, as the PR generated multisite is not available.  🤷‍♀️

PR #54, opened 2018-06-04
SG-114 media/groups

arlina-espinoza:
Adds the ability to add media entities to groups/departments, and sets up their permissions.


josh-chou:
Can you expand the tests to cover the new functionality?  Unless you think it's already covered.
Merge requests are much more interesting when reviewing code!  Thanks!!

PR #59, opened 2018-06-07
Sg 172 mandrill

josh-chou:
A few minor edits to the README.
Installed Mandrill module for outgoing emails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
   * pantheon machine token (https://pantheon.io/docs/machine-tokens/)  
 
 2. [Init Lando with Github](https://docs.devwithlando.io/cli/init.html#github)  
-`lando init github --recipe=pantheon`
+`mkdir sfgov && cd sfgov && lando init github --recipe=pantheon`
 3. `composer install`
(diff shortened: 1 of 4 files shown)


config/language.negotiation.yml
```diff
@@ -7,6 +7,7 @@ url:
     es: es
     fil: fil
     zh-hant: zh-hant
+    '': null
```
jacine: I got this locally too.  Not sure why it's happening, but it doesn't break anything AFAICT, so I think we can ignore it for now.

README.md
```diff
@@ -31,7 +31,7 @@ TLDR version:
 5. Delete branch
 
 ## Adding a contrib module
-1. Create a new branch `git clone -b new_branch`
+1. Create a new branch `git checkout -b new_branch`
 2. Install module with composer `composer require drupal/paragraphs`
```
jacine: I would add a step here (and we can do that later).  `composer update --lock`.   If that's not done, code like we have in #57 is around for the next person who updates.

PR #65, opened 2018-06-08
[SG-67], [SG-61]: Conditionally display cost fields, CSS tweaks.

jacine:
Brings in the `sfgov_admin` module which contains the CSS and Cost Paragraphs tweaks.


web/modules/custom/sfgov_admin/build/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+config.local.jsoon
```
josh-chou: .json?

jacine: OMG! Good catch @josh-chou.  LOL. Fix coming.

jacine:
I don't think much, if any of it would be related to the pattern library.  The admin area is a completely separate theme, and unless we're making a custom admin theme (which would be a massive undertaking) there won't really be much opportunity to share rendering/assets because they use entirely different templates and for the most part, completely different rendering pipelines.

PR #69, opened 2018-06-13
install config_split and set it so that mandrill configs do not get e…

josh-chou:
…xported


composer.json
```diff
@@ -39,6 +39,7 @@
         "drupal/components": "^1.0",
         "drupal/config_direct_save": "^1.0",
         "drupal/config_installer": "^1.0",
+        "drupal/config_split": "^1.3",
```
jacine: Good idea!  I completely forgot about config_split!  👍

PR #97, opened 2018-06-26
Sg 28 QA

isramv:
QA for footer and header.


jacine:
Looks like this failed because of unrelated issues that CircleCI AND Pantheon are currently having.

PR #110, opened 2018-06-28
Sg 181 workflow permissions

Jaesin:
This lets workflows work under group context. i.e.: `group/1/content/create/group_node:topic`


Jaesin:
It looks like some type of pantheon error. I don't have permission to re-run the test suite.

PR #115, opened 2018-06-29
Sg235 search view fields

josh-chou:


jacine:
Sorry for the delay on this. I forgot this was here!  I want to actually take a look tomorrow because I actually made a view mode for search results, I'm not sure if you are aware of that, or if that's what this is about.  I'll check back tomorrow.

PR #137, opened 2018-07-17
update ga id to use one created by ds

aekong:


config/google_analytics.settings.yml
```diff
@@ -1,4 +1,4 @@
-account: UA-120925231-1
+account: UA-122224372-11
```
jszwedko: Is this in the digital services GA account? I'm not seeing it as a property.

jszwedko: Or rather, did we create a new one for this site? Is it possible to just add it as a property to the existing GA account (102831134)? cc/ @hshaosf who set up the general Digital Services GA account.

aekong: yes, a new one was created.  i didn't know we had one already 😨

hshaosf: I think there are a couple of GA accounts. There is also one that's being used for the D7 sites.

hshaosf: probably outside of this PR, but i would recommend taking a look at implementing Google Tag Manager (https://www.google.com/analytics/tag-manager/ ) versus vanilla GA.

jszwedko: @aekong I added you to the existing DS GA account.

It'd be nice to have them all in one place (imo).

aekong: what if this is just for dev/testing?

aekong: going to look into ga tag manager per @hshaosf's suggestion

jszwedko: +1 for Google Tag Manager. Either way I think further proliferation of Google Analytics accounts makes it harder to keep track of and administer them.

aekong:
closing this pull request per discussions above

PR #141, opened 2018-07-19
uninstall google analytics, use google tag manager module instead

aekong:


config/google_tag.settings.yml
```diff
@@ -0,0 +1,25 @@
+container_id: GTM-NQB74NJ
```
jszwedko: 👏

PR #142, opened 2018-07-20
add pre-prod redirect module

henryjiang-sfgov:


henryjiang-sfgov:
Added redirect_anonymous and devel_vardump

composer.json
```diff
@@ -74,6 +75,7 @@
         "pantheon-systems/quicksilver-pushback": "~1",
         "rvtraveller/qs-composer-installer": "^1.1",
         "sf-digital-services/sfgov-pattern-lab": "dev-master#0.1",
+        "symfony/var-dumper": "^3.4",
```
jszwedko: Should this only be a development dependency?

PR #144, opened 2018-07-20
add anonymous redirect

henryjiang-sfgov:


jszwedko:
This seems to include a lot more changes than just the redirect, was that intentional?

henryjiang-sfgov:
I'm installing a contrib module which will redirect anonymous users on live-pantheonsite to sfgov.org.  The README calls for config export that exported all of my local configs.  I'm going to create another PR to see if I can get away without config export.

henryjiang-sfgov:
failed build, rejecting PR

PR #146, opened 2018-07-20
add anonymous_redirect, exclude configs

henryjiang-sfgov:


henryjiang-sfgov:
failed build, rejecting PR

PR #151, opened 2018-07-24
SG-229, SG-269. Add sidebar missing information to Departments and Co…

henryjiang-sfgov:
SG-229, SG-269. Add sidebar missing information to Departments and Contact us page


composer.json
```diff
@@ -45,13 +45,15 @@
         "drupal/console": "^1",
         "drupal/core": "^8",
         "drupal/ctools": "^3.0",
+        "drupal/devel": "^1.2",
```
jszwedko: I feel like this and `kint` should be in `require-dev` since they aren't required in production.

henryjiang-sfgov:
This PR has too many stuff in it. Pulling it back.

PR #154, opened 2018-07-25
fix side bar for contact us and department pages

henryjiang-sfgov:


composer.json
```diff
@@ -45,6 +45,7 @@
         "drupal/console": "^1",
         "drupal/core": "^8",
         "drupal/ctools": "^3.0",
+        "drupal/devel": "^1.2",
```
jszwedko: Did you mean to move this up? I still expected to see this below in `require-dev`.

henryjiang-sfgov: I'm not sure where this belongs. If I move it under require-dev, CI build breaks.

jszwedko: Ah, yes, I assume you mean this build https://circleci.com/gh/SFDigitalServices/sfgov/934 ?

That looks to be because we have the `devel` module enabled on Pantheon. I guess the question is, do we want that?

aekong: might need to update composer.lock and remove associated yml files? i had to do that for google tag manager.  someone more composer savvy pls chime in ❤️

hshaosf: probably want to disable *devel* module in Drupal first, then we can move it to *require-dev* in composer

hshaosf: in general, just be more cautious of doing a straight configuration dump from development machine and checking it in without reviewing what was being exported. that's probably why the *devel* module was enabled on Pantheon in the first place.

henryjiang-sfgov: yeah I wish drush export allows you to exclude configurations.

hshaosf: https://www.drupal.org/docs/8/modules/configuration-split/creating-a-simple-split-configuration-dev-modules-only-in-dev

jszwedko: 👍 to configuration split module; I believe it's already in use in this project.

There are also some alternative approaches mentioned on: https://drupal.stackexchange.com/questions/188817/how-to-prevent-devel-module-to-be-installed-on-production-enviroments

henryjiang-sfgov: I like the drush method, since we're already using drush as part of the deployment process

aekong: just curious - is there a reason why the paragraph module was updated in the same pr as the side bar fix? was the paragraph module update required to implement the side bar fix? my preference would be to separate feature/fix branches from module update branches, if possible.  then again, i'm still working on my drupal literacy, so ¯\_(ツ)_/¯

jszwedko: 👍 if it is necessary, a note in the description of the PR would help readers understand (also serves as a record for people to look back to). Additionally, putting that info in the commit message is always helpful too.

henryjiang-sfgov: I would rather separate this PR out, the original PR did not pass test because of the Devel module. I had to fix that, but in the process, the module Paragraphs need to be updated.

jszwedko: 👍 I don't mind module updates being in PRs like this, but it'd still be nice to know more of the "why". I.e. why did the paragraphs module need to be updated?

PR #156, opened 2018-07-25
Pin sfgov-pattern-lab to the master branch

jszwedko:
My understanding is that we were continuously retagging `sfgov-pattern-lab` as `0.1` every time we updated master. This just points to the `master` branch.

I think this is preferable as:

* it simplifies the workflow during this initial rapid development push
* git tags are meant to be immutable

Eventually we will likely want to start tagging official releases of the pattern library using semantic versioning or similar, especially as more projects begin to depend on it, but this seems sensible for the time being.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -74,7 +74,7 @@
         "oomphinc/composer-installers-extender": "^1.1",
         "pantheon-systems/quicksilver-pushback": "~1",
         "rvtraveller/qs-composer-installer": "^1.1",
-        "sf-digital-services/sfgov-pattern-lab": "dev-master#0.1",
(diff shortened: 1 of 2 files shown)


aekong:
Doesn't this also mean that if something broken is checked in to master that it would break on the sites relying on this pattern library? I figure with forcibly moving the tag there's a little more control.  Thoughts on this? @jszwedko @henryjiang-sfgov @hshaosf

aekong:
sorry i read too fast - saw your "seems sensible for the time being" 👍

jszwedko:
Yeah, it seems like at this point we are just assuming master is stable, but it would be good to start tagging releases in the future.

PR #158, opened 2018-07-26
Sg 186 search results

aekong:
- match the container on other pages
- fix the search bar color
- bring back the js to prevent click on smaller screen size
- display first phone number in department search result, if it exists


web/themes/custom/sfgovpl/src/js/search.js
```diff
@@ -14,8 +14,10 @@
       // Search Toggle.
 
       $('header .block-views-exposed-filter-blocksearch-page-1 .form-submit').on('click', function(e) {
-        e.preventDefault();
-        $('.responsive-search--container').toggle();
+        if($(window).width() < 770) {
```
jszwedko: Could this number be pulled out as a constant somewhere?

aekong: It _could_

jszwedko: _but will it?_

henryjiang-sfgov: These type of behavior is pretty common throughout other areas.  Let's create a Jira ticket to refactor them.  Fixing it here would just be an one-off.

henryjiang-sfgov: SG-286 is created to keep track of this discussion.

jszwedko: @henryjiang-sfgov I'm not seeing any other examples of hard coding a width in the Javascript (using `ack '\.width'`) in the theme. Do you have a link for comparison?

henryjiang-sfgov: @jszwedko src/js/resizeToolbarDisable.js

jszwedko: Mmm, yes, I see, I missed that one. I still think we should pull them out sooner than later rather than continue this pattern.

PR #159, opened 2018-07-26
move devel module to require-dev

aekong:
make devel a dev dependency


jszwedko:
(to be clear, I'm more than happy to see them kept up-to-date)

aekong:
@jszwedko i did, it was, and i think it's ok (i hope).  any objections? @henryjiang-sfgov @josh-chou @hshaosf

jszwedko:
I'm all about it.

aekong:
#176 updated composer.lock as well, which was merged to master.  merged master to this branch, so all should be 👍 (hopefully)

PR #161, opened 2018-07-27
Various layout issues

aekong:
- fix missing margins on department and service landing pages
- fix search result missing phone icon
- fix dept display on transaction search result


web/themes/custom/sfgovpl/src/sass/_blocks.scss
```diff
@@ -235,19 +235,25 @@
   }
 }
 
-.contextual-region {
-  padding: 41px 0 21px 0;
-}
+// #block-sfgovpl-local-tasks.contextual-region {
```
jszwedko: Should we just delete this?

PR #162, opened 2018-07-27
add phone number field theme to remove label from display, fix bug in…

henryjiang-sfgov:
… event_subscriber code

### DIFF
--- a/web/modules/custom/sfgov_event_subscriber/src/EventSubscriber/RedirectEventSubscriber.php
+++ b/web/modules/custom/sfgov_event_subscriber/src/EventSubscriber/RedirectEventSubscriber.php
@@ -41,13 +41,11 @@ public function redirectBasedOnField(GetResponseEvent $event) {
         $node = $event->getRequest()->attributes->get('node'); 
       
         if($node && $node->hasField('field_direct_external_url') ){
-          $field_external_url = $node->get('field_direct_external_url');
-          $external_url = $field_external_url->getValue()[0]['uri'];
+          $field_external_url = $node->get('field_direct_external_url')->getValue();
 
-          if($external_url != '' ){  
+          if( !empty($field_external_url[0]) && $field_external_url[0]['uri'] != ''){
               // This is where you set the destination.
-              //$redirect_url = Url::fromUri($external_url->getValue());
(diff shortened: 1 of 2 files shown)


web/modules/custom/sfgov_event_subscriber/src/EventSubscriber/RedirectEventSubscriber.php
```diff
@@ -41,13 +41,11 @@ public function redirectBasedOnField(GetResponseEvent $event) {
         $node = $event->getRequest()->attributes->get('node'); 
       
         if($node && $node->hasField('field_direct_external_url') ){
-          $field_external_url = $node->get('field_direct_external_url');
-          $external_url = $field_external_url->getValue()[0]['uri'];
+          $field_external_url = $node->get('field_direct_external_url')->getValue();
 
-          if($external_url != '' ){  
+          if( !empty($field_external_url[0]) && $field_external_url[0]['uri'] != ''){
```
jszwedko: Don't change this now, but I believe, since `empty` is a language construct rather than a function, you can _just_ do `empty($field_external_url[0]['uri'])`.

henryjiang-sfgov: can't just do empty($field_external_url[0]['uri']), it'll return true if uri is an empty string.

jszwedko: Sorry, I meant `!empty($field_external_url[0]['uri']`

jszwedko: Right, I'm saying `empty` doesn't do that, see: http://codepad.org/IXPM9Epg

Again, not a big deal 😄

henryjiang-sfgov: aha good point, I was too focused on checking for error.  Empty indeed is a PHP language construct, it'll fail the test and does nothing(?my guesswork) rather than throwing an error.

PR #167, opened 2018-07-30
Transaction page commits

henryjiang-sfgov:


web/modules/custom/sfgov_event_subscriber/src/EventSubscriber/RedirectEventSubscriber.php
```diff
@@ -30,7 +28,12 @@ public static function getSubscribedEvents() {
     return $events;
   }
 
-public function redirectBasedOnField(GetResponseEvent $event) {
+  public function redirectBasedOnField(GetResponseEvent $event) {
```
jszwedko: Same comment as https://github.com/SFDigitalServices/sfgov/pull/166/files#r206236555

josh-chou: I created a separate Jira task for this so it can be separated out from the transaction theming task.
https://sfgovdt.jira.com/projects/SG/issues/SG-308

PR #173, opened 2018-08-01
Sg 246 mobile header

aekong:
- fix translate odd zoom behavior
- fix mobile menu display
- fix city seal height
- link sfgov.org in new site alpha banner


config/simple_block.simple_block.alpha_banner_block.yml
```diff
@@ -7,5 +7,5 @@ dependencies:
 id: alpha_banner_block
 title: 'Alpha Banner Block'
 content:
-  value: "<p>This site is in progress. Visit <a href=\"#\">sfgov.org</a> for the current.</p>\r\n"
+  value: "<p>This site is in progress. Visit <a href=\"http://sfgov.org\" target=\"_blank\">sfgov.org</a> for the current.</p>\r\n"
```
jszwedko: Might make this `https://` to encourage 🔒s everywhere

aekong: i happened to be looking for another commit to justify a ci build, so thx 👍

web/themes/custom/sfgovpl/src/sass/_blocks.scss
```diff
@@ -63,27 +66,17 @@
 .gtranslate {
   cursor: pointer;
   display: flex;
-  max-width: 150px;
+  max-width: 155px;
```
jszwedko: On Firefox the `e` seems to be being cut off still. 160px seems to do it, though I imagine it's probably related to the font width so there may be a better way to size it that will automatically expand depending on the width of the text.

screen shot 2018-08-02 at 1 31 08 pm

PR #177, opened 2018-08-02
Sg 246 mobile header

aekong:
- https on sfgov.org link in alpha banner
- cursor pointer on translate "button"


config/simple_block.simple_block.alpha_banner_block.yml
```diff
@@ -7,5 +7,5 @@ dependencies:
 id: alpha_banner_block
 title: 'Alpha Banner Block'
 content:
-  value: "<p>This site is in progress. Visit <a href=\"http://sfgov.org\" target=\"_blank\">sfgov.org</a> for the current.</p>\r\n"
+  value: "<p>This site is in progress. Visit <a href=\"https://sfgov.org\" target=\"_blank\">sfgov.org</a> for the current.</p>\r\n"
```
jszwedko: ❤️

web/themes/custom/sfgovpl/src/sass/_blocks.scss
```diff
@@ -100,6 +100,7 @@
     right: 40px;
     padding: 5px 5px 5px 50px;
     background: transparent;
+    cursor: pointer;
```
jszwedko: Not a question that needs to be answered here, but this is making me wonder how much of this should actually be in the pattern library.

aekong: most of it! MVP!

PR #179, opened 2018-08-02
Sg252 transaction

josh-chou:


web/themes/custom/sfgovpl/src/sass/_base.scss
```diff
@@ -4,3 +4,10 @@
 @import '../../../../../libraries/sfgov-pattern-lab/pattern-lab/source/_patterns/00-protons/02-mixins';
 @import '../../../../../libraries/sfgov-pattern-lab/pattern-lab/source/_patterns/00-protons/01-grid.scss';
 @import 'mixins';
+
+body {
```
jszwedko: Should this go in the Pattern Library?

Should I be asking this question?

josh-chou: hopefully this question can be answered at the meeting today.
but i decided to put these changes here to minimize the scope.

PR #180, opened 2018-08-03
twig function file_url does not work in pattern lab twig file, move t…

henryjiang-sfgov:
…he function to template

### DIFF
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--document.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--document.html.twig
@@ -55,7 +55,8 @@ not paragraph.isPublished() ? 'paragraph--unpublished',
         {% block content %}
             {% include "@atoms/09-document/09-document.twig" with {
             "parent_bundle": parent.bundle(),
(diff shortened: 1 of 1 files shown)


web/themes/custom/sfgovpl/templates/paragraphs/paragraph--document.html.twig
```diff
@@ -55,7 +55,8 @@ not paragraph.isPublished() ? 'paragraph--unpublished',
         {% block content %}
             {% include "@atoms/09-document/09-document.twig" with {
             "parent_bundle": parent.bundle(),
-            "field_file": paragraph.field_file
+            "field_file": paragraph.field_file,
+            "fileurl": file_url(paragraph.field_file.entity.field_media_file.entity.uri.value)
```
aekong: i saw that problem too.  weird that the pattern lab twig file won't recognize file_url, but it is recognized in theme twig file.  what gives?

henryjiang-sfgov: yeah, I tried updating twig in the pattern lab, but some of the other twig extension relies on older version of twig.

PR #181, opened 2018-08-03
Fix address sidebar

henryjiang-sfgov:


henryjiang-sfgov:
doing another build, need to format the address field as the country name is showing up

PR #188, opened 2018-08-08
Sg 246 mobile header

aekong:
adjust mobile header for even smaller screens


web/themes/custom/sfgovpl/src/sass/_blocks.scss
```diff
@@ -71,6 +71,10 @@
     width: 41px;
     margin-right: 10px;
   }
+  @include media-max($tiny-screen) {
```
hshaosf: > $tiny-screen

 @jszwedko would appreciate this. no more hardcoding constants.

aekong: wwjd!

PR #189, opened 2018-08-08
indent depts for topic pages, remove bottom border for last dept in list

aekong:


web/themes/custom/sfgovpl/src/sass/_views.scss
```diff
@@ -72,23 +72,31 @@ article.transaction .sfgov-container--content-right {
   }
 }
 
-.view-services-block-1 {
+.view-services-block-1,
+.view--departments-block-1 {
   padding: 0;
   @include media($medium-screen) {
-    padding: 61px 0 41px 0;
+    padding: 60px 0 40px 0;
```
jszwedko: 👏  the idea of previously being so exact with the number of pixels gives me hives.

web/themes/custom/sfgovpl/src/sass/_views.scss
```diff
@@ -72,23 +72,31 @@ article.transaction .sfgov-container--content-right {
...
-      .sfgov-transaction-search--container {
+      .sfgov-transaction-search--container,
+      .department-search-result--container {
         border-bottom: 0;
       }
     }
   }
 }
 
+.view--departments-block-1 {
```
jszwedko: This reads to me that the first block on this view, whatever it happens to be, it should have this padding. Is that accurate? I'm mostly wondering if it'd make sense to give the block a more semantic name.

aekong: Looks like the class name is coming from https://github.com/SFDigitalServices/sfgov/blob/master/web/themes/custom/sfgovpl/templates/views/views-view--departments--block-2.html.twig.  So I _suppose_ I _could_ make the class name a bit more semantic, but the twig file name suggestion remains the same.

Also, I would think the class name would be `view--departments-block-2` given this is block 2 defined in the Departments view.

web/themes/custom/sfgovpl/src/sass/_views.scss
aekong: Yeah, it was weird to me.  I'm pretty sure these were coming from inspect mode on the invision mockups - which seems to be invision trying to intelligently determine the spacing between elements in a mockup

web/themes/custom/sfgovpl/src/sass/_views.scss
jszwedko: It's probably my ignorance of the content structure. Is block 2 special? Or is whatever block 2 happens to be on the department's view need this styling. I.e. could I reorder the blocks and the style would still be accurate?

jszwedko: (also my ignorance of Drupal)

aekong:
changing the order of the blocks does not appear to affect the machine name (`block_2`)

jszwedko:
Hmm, interesting. I might recommend using more semantic names if possible then, but don't let that hold this up. Just something to think about.

hshaosf:
looks like the class names were hardcoded into the twig file: 
https://github.com/SFDigitalServices/sfgov/blob/master/web/themes/custom/sfgovpl/templates/views/views-view--departments--block-1.html.twig#L34-L37
https://github.com/SFDigitalServices/sfgov/blob/master/web/themes/custom/sfgovpl/templates/views/views-view--departments--block-2.html.twig#L34-L37

aekong:
just tried to change the machine name to something more meaningful and drupal was no longer using the right template suggestion.  someone with more drupal skillz pls contribute to the success of this pr

PR #192, opened 2018-08-14
Sg 323 tx external urls

aekong:
doing this in the theme/template layer

also requires an update to pattern library as the pattern library has to accommodate a hash now - the calling template passes in a url and boolean to determine if the link should open in a new window or not (based on logged in user)

open to suggestions on a better approach

this is the diff to review: https://github.com/SFDigitalServices/sfgov/pull/192/files#diff-e8b47a90240e6702bc9b47527544a322


aekong:
@hshaosf says we should handle the logic in a module and send the variables to the template.  closing for now to revisit locally

PR #194, opened 2018-08-14
Update package.json in sfgovpl

jszwedko:
Current version of hoek has a vulnerability. Updating actually seems to remove it as a dependency.

Reran gulp after this update and nothing seemed to change.


jszwedko:
Yep! I just ran `npm update`

PR #198, opened 2018-08-16
local behat testing

aekong:
- need to commit .lando.yml to do this
- need to prevent https redirect on local development environment (get nested calls exception otherwise)


web/sites/default/settings.php
```diff
@@ -53,30 +53,29 @@
  * example.org, with all subdomains included.
  */
 
-
 // force https on sf.gov
 if (isset($_ENV['PANTHEON_ENVIRONMENT']) && php_sapi_name() != 'cli') {
+  if($_ENV['PANTHEON_ENVIRONMENT'] != 'lando') { // lando is local env
```
aekong: is this check ok? or is there something more robust?

web/sites/default/settings.php
```diff
@@ -53,30 +53,29 @@
...
     // Redirect to HTTPS on every Pantheon environment.
     $primary_domain = $_SERVER['HTTP_HOST'];
-
     if ($_SERVER['HTTP_HOST'] != $primary_domain
```
aekong: this condition probably doesn't need to be here, since the line above is `$primary_domain = $_SERVER['HTTP_HOST'];`

.gitignore
```diff
@@ -13,7 +13,6 @@
 /web/private/scripts/quicksilver
 
 # Ignore scaffold files
-.lando.yml
```
jszwedko: 👍 I'm surprised we weren't already checking this in

web/sites/default/settings.php
jszwedko: Yeah, this is weird.

web/sites/default/settings.php
jszwedko: I might fold this into the conditional above it to avoid the nesting.

jszwedko:
👏

web/sites/default/settings.php
aekong: thought i needed that `isset($_ENV['PANTHEON_ENVIRONMENT'])`

jszwedko: I _think_ you do too. I was thinking something like:

```
 if (isset($_ENV['PANTHEON_ENVIRONMENT']) && $_ENV['PANTHEON_ENVIRONMENT'] != 'lando'
     && php_sapi_name() != 'cli') { 
```

aekong: 👍

aekong:
THANK YOU

PR #201, opened 2018-08-21
target=_blank for external links

josh-chou:


josh-chou:
I tried to refactor the code but if you look closely, it is not as repetitive as it looks.  Different classes are being set at different levels depending on the conditions.

aekong:
Yeah I guess I was thinking that if a label would never display or if there are never multiple we could eliminate those conditions.  But I suppose those situations probably would happen

________________________________

PR #217, opened 2018-09-06
SG-243 install extlink module and configure external links to open in…

josh-chou:
… new window

### DIFF
--- a/config/extlink.settings.yml
+++ b/config/extlink.settings.yml
@@ -1,0 +1,18 @@
+extlink_target: true
+extlink_target_no_override: false
+extlink_nofollow: false
+extlink_follow_no_override: false
+extlink_subdomains: true
+extlink_alert: false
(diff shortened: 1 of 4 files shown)


config/extlink.settings.yml
```diff
@@ -0,0 +1,18 @@
+extlink_target: true
+extlink_target_no_override: false
+extlink_nofollow: false
+extlink_follow_no_override: false
+extlink_subdomains: true
+extlink_alert: false
+extlink_alert_text: 'This link will take you to an external web site. We are not responsible for their content.'
```
aekong: why do i like this alert text so much 👍 

just fyi, content will probably want to change this.  i think it's super boss, but 🤷‍♂️

jszwedko: `Visit at your own risk`

PR #220, opened 2018-09-10
sg-349 alignment issues on tablet/mobile for dept/tx pages

aekong:


aekong:
they were from the pattern library, but the theme sass references the pattern lib sass, so had to gulp generate the dist

PR #223, opened 2018-09-12
SG-243 render phone numbers as links and enforce xxx-xxx-xxxx format

josh-chou:


web/modules/custom/sfgov_admin/js/src/sfgov-admin.js
```diff
@@ -91,4 +91,20 @@
...
 
+  // Enforce telephone numbers have format XXX-XXX-XXXX
+  Drupal.behaviors.sfgovFormatTelephone = {
+    attach: function(context, settings) {
+      $('#edit-submit').once().on('click', function(e) {
+        var phoneField = $('.form-tel');
+        if (phoneField.length > 0) {
+          var regex = new RegExp("\\d+", "g");
+          var number = phoneField.val().match(regex).join("");
+          if (number.length == 10) {
```
aekong: what do we do in situations where the phone number doesn't have the correct number of characters? are we enforcing this through drupal manage fields?

web/modules/custom/sfgov_admin/js/src/sfgov-admin.js
```diff
@@ -91,4 +91,20 @@
     }
   }
 
+  // Enforce telephone numbers have format XXX-XXX-XXXX
+  Drupal.behaviors.sfgovFormatTelephone = {
+    attach: function(context, settings) {
+      $('#edit-submit').once().on('click', function(e) {
+        var phoneField = $('.form-tel');
+        if (phoneField.length > 0) {
+          var regex = new RegExp("\\d+", "g");
```
aekong: noice

web/modules/custom/sfgov_admin/js/src/sfgov-admin.js
```diff
@@ -91,4 +91,20 @@
     }
   }
 
+  // Enforce telephone numbers have format XXX-XXX-XXXX
+  Drupal.behaviors.sfgovFormatTelephone = {
+    attach: function(context, settings) {
+      $('#edit-submit').once().on('click', function(e) {
+        var phoneField = $('.form-tel');
```
aekong: is this a class added by drupal when rendering the field? is it the same class in both creating and editing? I think drupal names fields according to whether or not the user is creating vs. editing.  So something like "field_phone_num" in node create form could be "edit_field_phone_num" in node-edit form.  I _think_.  Sorry for this comment if I'm mistaken

web/modules/custom/sfgov_admin/js/src/sfgov-admin.js
josh-chou: If the phone number doesn't have 10 digits, the new code doesn't do anything.  In other words, it just continues to do what it is doing now - which looks like it just displays what the editor inputs.  There doesn't appear to be any validation.

web/modules/custom/sfgov_admin/js/src/sfgov-admin.js
josh-chou: I checked edit and create.  The classname is the same.  And the submit button id is the same also.
Good call though.  I only tested in edit mode during development.

aekong: Great thanks for checking

PR #226, opened 2018-09-13
Sg 356 person page

aekong:
basic styling for the person page

### DIFF
--- a/web/modules/custom/sfgov_event_subscriber/src/EventSubscriber/RedirectEventSubscriber.php
+++ b/web/modules/custom/sfgov_event_subscriber/src/EventSubscriber/RedirectEventSubscriber.php
@@ -18,42 +18,34 @@ use Drupal\Core\Session\AccountInterface;
  */
 class RedirectEventSubscriber implements EventSubscriberInterface {
 
-    /**
+  /**
    * {@inheritdoc}
    */
   public static function getSubscribedEvents() {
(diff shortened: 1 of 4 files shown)


web/modules/custom/sfgov_event_subscriber/src/EventSubscriber/RedirectEventSubscriber.php
```diff
@@ -18,42 +18,34 @@
...
-      // don't redirect logged in users.
-      $account = \Drupal::currentUser();
-      if ( !empty($account->id()) ) {
-          return;
-      }
-    // only redirect if in view mode. This relies on proper URL path setting for all nodes.  
-    $uri_array = explode('/',  $event->getRequest()->getRequestUri() ); 
-    if($uri_array[1] == 'node' && ( $uri_array[2] && is_numeric($uri_array[2]) ) && $uri_array[3] != '' ){
-        // do nothing
-        //\Drupal::logger('sfgov_event_subscriber')->notice('edit mode');
```
josh-chou: Is this the case when you are signed in, it won't redirect?

aekong: yeah, so that additional empty conditional is unnecessary, since there's already a check.  henry already removed in a local branch, but it didn't get checked in.  coordinated offline, so those changes are here now

PR #238, opened 2018-10-18
Sg 369 custom meta tags

aekong:


web/themes/custom/sfgovpl/sfgovpl.theme
```diff
@@ -19,33 +19,36 @@ function sfgovpl_preprocess_html(&$variables) {
...
-      $phone = Paragraph::load($paragraphPhoneId);
-      
-      $paragraphAddressId = $node->get('field_address')->getValue()[0]['target_id'];
-      $paragraphAddress = Paragraph::load($paragraphAddressId);
-
-      $address = $node->get('field_address')->referencedEntities()[0];
-      
-      $nid = $node->id();
-      $dept = [
-        'id' => $nid,
```
josh-chou: it looks like we are losing address2, postal_code, locality, and organization.
are these fields unnecessary?

aekong: we really only need address and phone, everything else was a nice to have when i put a json string in the content.  but decided against json in the content attr

PR #253, opened 2019-02-09
Gtranslate intercept

aekong:
- choose between drupal added translation and gtranslate
- do not show machine vs human translated notification


aekong:
> is this testable?
> would be nice if there was test coverage since this seems like functionality that could easily break.

Yes it is! I'll write one 👍

PR #257, opened 2019-02-26
Sg 486

aekong:
* add checks for existence of related dept and related topic
* move hooks from .theme to includes folder in appropriate .inc files to keep with set standard


aekong:
thank you!

PR #263, opened 2019-03-18
Sg 498 search autocomplete

aekong:
* implement autocomplete
* move top search suggestions to search module


web/modules/custom/sfgov_search/js/sfgov_search.js
```diff
@@ -17,6 +21,78 @@ function Search311() {
...
+
+  this.topSearchSuggestions = function() {
+    var topSearchSuggsSelector = '.sfgov-top-search-suggestion';
+    var topSearchSuggs = $(topSearchSuggsSelector);
+    if(topSearchSuggs.length > 0) {
+      
+      var containerId = this.topSearchContainerSelector.replace('#', '');
+      var containerSelector = '#' + containerId;
+      
+      $(topSearchSuggs[0]).before('<div id="' + containerId + '"><h4>Top searches:</h4></div>');
```
josh-chou: translation

web/modules/custom/sfgov_search/js/sfgov_search.js
```diff
@@ -17,6 +21,78 @@ function Search311() {
     }
   };
 
+  $('body').click(function(e) {
+    var clickTarget = $(e.target);
+    if(clickTarget.attr('id') != 'edit-sfgov-search-input') {
+      $(_this.autocompleteContainerSelector).hide();
```
josh-chou: where is _this defined?  Is it referencing something different than `this`?

aekong: `_this` is a class reference defined at the end of the class definition.  Had to do it because in the scope above, `this` would refer to the dom element created by the anon callback in the click event.

PR #266, opened 2019-04-03
Sg 423 dept homepage

aekong:


web/themes/custom/sfgovpl/templates/node/node--event--full.html.twig
```diff
@@ -0,0 +1,42 @@
...
+            {{ content.field_start_date }}<br/>
+            {{ content.field_end_date }}
+          </div>
+          <div class="event--meta-cost">
+            {{ content.field_cost }}
+          </div>
+        </div>
+        <div class="event--meta-right">
+        <pre>
+          {{ dump(content.field_address) }}
```
zakiya: Should remove dump statement if people will be creating events in this iteration.

aekong: DOH!

PR #268, opened 2019-05-03
News

zakiya:
Culmination of work for SG- 532,533, & 534

Dev notes:
- Added config_exclude so config exports could be cleaner on local. To use, enable locally and add `$settings['config_exclude_modules'] = ['devel', 'stage_file_proxy', 'config_exclude', 'kint', 'search_kint', 'libraries_debug', 'entity_clone', 'reroute_email', 'devel_generate'];
` or similar to settings.local.php

- Added views_infinite_scroll. Reusing code from the search infinite scroll would take significantly more time to develop since that the search was custom development.

- I used existing variables and mixins from either pattern lab or department homepages where clear but I also added a component folder to the the theme so new components can be developed there. I also started a font-style folder base on the style guide produced by Zeplin which I will share/demo later. 

- There are a few items that problems/questions that need whole team attention that I will detail in Jira tickets.


web/themes/custom/sfgovpl/src/sass/view/_view-news-embed.scss
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  .views-row {
+    @include media($medium-screen) {
+      float: left;
+      margin-right: 5%;
+      width: 30%;
+    }
+
+    &:nth-child(3) {
```
aekong: is doing stuff like this (`nth-child(3)`) avoidable? or is it just what we get when we let drupal do the rendering?

zakiya: Is there a reason you want to avoid :nth-child? :last-child can be used of course. Views doesn't output a first or last class by default so fixing with css is fastest.

PR #270, opened 2019-05-14
redirect when dept has go direct to website checked

aekong:


web/modules/custom/sfgov_event_subscriber/src/EventSubscriber/RedirectEventSubscriber.php
```diff
@@ -36,16 +36,30 @@ public function redirectBasedOnField(GetResponseEvent $event) {
...
     $node = $event->getRequest()->attributes->get('node'); 
-
-    if($node && $node->hasField('field_direct_external_url') ){
-      $field_external_url = $node->get('field_direct_external_url')->getValue();
-
-      if( !empty($field_external_url[0]) && $field_external_url[0]['uri'] != ''){
-        // This is where you set the destination.
-        $response = new TrustedRedirectResponse($field_external_url[0]['uri']);
-        $event->setResponse($response);
+    if($node && $node->isPublished()) {
```
nicolesque: I'm just curious, why do we check for the external URL before we check for the boolean? Might it be a bit faster to check to see if they want to redirect before we even bother with the other field?

My JS is rusty so also maybe I am misreading the order here...

aekong: no worries - `field_direct_external_url` is a field used by transactions and person content types.  that bit of code (lines 40-49) checks for the existence of this field on any content type.  Unfortunately, this field does not exist on the department content type.  When dept content type was built, it used `field_dept_url` in conjunction with the `field_go_to_current_url`checkbox to determine whether or a dept node should redirect (which was not implemented and the reason for this pr).  I'm not sure why this combination of fields was used as opposed to reusing `field_direct_external_url`, though.  🤷‍♂

PR #279, opened 2019-05-22
fix search header padding

aekong:


aekong:
@nicolesque 

broken: https://dev-sfgov.pantheonsite.io/search?keyword=birth

fixed with this pr: https://ci-2369-sfgov.pantheonsite.io/search?keyword=birth

PR #289, opened 2019-05-28
579 basic

zakiya:


web/themes/custom/sfgovpl/src/sass/_mixins.scss
```diff
@@ -40,6 +40,14 @@
   }
 }
 
+@mixin columns-720($left-or-right) {
+  @include media($medium-screen){
+    float: #{$left-or-right};
+    max-width: 720px;
+    width: calc((100% - 20px)/3*2);
```
aekong: have you had a chance to look at how this behaves in IE 11? caniuse.com reports that nested expressions round differently

zakiya: Good question. I did not check before as I've never run into calc actually being a problem for IE compatibility. Checking just now at max screen width, IE 11 shows calculated width as 713.33 and Chrome as 713.328. There's a 20px empty space between the column mixins.  I think if we were using calc more aggressively or exactly it might be a problem.

aekong: cool, thanks for checking!

PR #295, opened 2019-05-31
SG-566 - Document and mailing address paragraph should not have been …

zakiya:
…added to field_help.


config/field.field.node.transaction.field_help.yml
```diff
@@ -51,8 +47,8 @@ settings:
         weight: 20
         enabled: false
       document:
-        enabled: true
         weight: 20
+        enabled: false
```
aekong: just curious - did you try deleting these fields? as opposed to disabling?

PR #299, opened 2019-06-07
594 404 page

zakiya:
I also updated the scss color values in this PR.


web/themes/custom/sfgovpl/src/js/search404.js
```diff
@@ -0,0 +1,21 @@
+(function ($) {
```
aekong: I believe this was already changed for all sfgov search block text inputs, is this not the case for the search block used for 404 pages?

zakiya: The 404 search box is not but I don't see the updated text on https://sf.gov homepage or https://sf.gov/search. It says "Search" at all screen sizes. I assume it doesn't need to be applied in the header?

aekong: so turns out the figma designs were incorrect, the search input placeholder should just be "search".  the commit f94f2e9381a3146ef2b29b70a2164b5a31ae8339 resolves this convo

PR #325, opened 2019-06-26
Drush update

aekong:


aekong:
> Does ‘lando pull’ of only code and files work without error? Only need to call lando getdb if needing the db?
we don't usually pull code down from pantheon, but yeah, pulling files is 👍 .  `lando pull`-ing the db results in this error: 
```
Fatal error: Class 'Drush\Commands\DrushCommands' not found in /etc/drush/drupal-8-drush-commandfiles/Commands/site-audit-tool/SiteAuditCommands.php on line 18
Ok
Drush command terminated abnormally due to an unrecoverable error. 
```

PR #329, opened 2019-07-02
SG-649 - Theme in person address on transaction and remove extra logic.

zakiya:


web/themes/custom/sfgovpl/templates/paragraphs/paragraph--in-person-location.html.twig
```diff
@@ -52,36 +52,9 @@ set classes = [
...
-            "address_line2": field_address.address_line2,
-            "dependent_locality": field_address.dependent_locality,
-            "locality": field_address.locality,
-            "postal_code": field_address.postal_code,
-            "administrative_area": field_address.administrative_area
-          } %}
-        {% endif %}
-      {% endif %}
-      <div class="address-in-paragraph">
         {{ content }}
```
aekong: just curious - if later we want to re-integrate this with a component from the pattern library, would we need to break this all out again and pass the values to the component? or will we need to drill down further into the template hierarchy?

zakiya: I wouldn't implement this this way at all. Unless you were getting data from sources outside of drupal, you don't need this level of control when the values are just getting displayed. Any custom work can be done on field level and with display modes.

aekong: Good to know, it seems pretty verbose.  Makes sense to do it on the field level instead of adding all this logic in the template to pull out values.

PR #332, opened 2019-07-05
Sg 653 homepage cleanup

aekong:


web/themes/custom/sfgovpl/src/sass/_search.scss
```diff
@@ -0,0 +1,52 @@
...
+  clear: both;
+  position: absolute;
+  left: 8px;
+  top: 50px;
+  position: absolute;
+  z-index: 1000;
+  background: $c-white;
+  width: calc(100% - 156px);
+  padding: 0;
+  box-shadow: #00000040 0px 5px 5px 2px;
```
zakiya: Check this hex value?

aekong: thx - was using alphahex, updated to use rgba

PR #336, opened 2019-07-10
Alerts

zakiya:


config/language/zh-hant/views.view.block_content.yml
```diff
@@ -0,0 +1,26 @@
...
+            items_per_page_label: 每頁的項目數
+            items_per_page_options_all_label: '- 全部 -'
+            offset_label: 位移
+      fields:
+        changed:
+          label: 已更新
+        operations:
+          label: 操作
+  page_1:
+    display_title: 頁面
```
aekong: i'm not familiar with this, but it looks like you had a to add a translation for every language? was this to get around that block bug?

config/block.block.alertblock.yml
```diff
@@ -17,7 +17,6 @@ settings:
   label: 'Alert block'
   provider: sfgov_alerts
   label_display: '0'
-  number: '5'
-  alert_text: 'this text has changed.'
-  expiration_date: ''
+  alert_expiration: '2025-05-05'
+  alert_text: 'test1265656<a href="/text">text</a>'
```
aekong: i suppose these are default values for either of these fields? not sure if there is a spec'ed default for expiration date (if indeed that's what this config is doing)

config/language/zh-hant/views.view.block_content.yml
aekong: comment no longer relevant

config/block.block.alertblock.yml
aekong: comment no longer relevant

aekong:
squash and merging as admin because i made a very small contribution to this repo and i can't approve myself

PR #348, opened 2019-07-31
601 hover states

zakiya:
As part of this, I refactored the service cards on the homepage, Topic pages, and departments to all use the same components and styles.


zakiya:
Pull request should be against develop not master. Closing.

PR #351, opened 2019-07-31
SG-678 Need to know link color.

zakiya:


zakiya:
Whomp Whomp. PR should have been against develop. Closing.

PR #358, opened 2019-08-09
Sg 619 formbuilder integration

josh-chou:


web/sites/default/settings.php
```diff
@@ -105,8 +105,9 @@
 
 /**
  * If there is a local settings file, then include it
+ * Local settings file need to be in sites/default/files/private for Pantheon sites
  */
-$local_settings = __DIR__ . "/settings.local.php";
+$local_settings = "sites/default/files/private/settings.local.php";
```
aekong: so this actually replaces the `settings.local.php` where i have additional settings, like not caching stuff so i don't have to `drush cr` whenever i make a css or js update.  is there anyway to expose the `$settings['formbuilder_include_url']` additively instead of replacing the entire `settings.local.php`?

josh-chou: Yes, we can introduce another settings file in the pantheon defined path, so that there are two settings.local.php files.  Or maybe instead of settings.local.php, rename it settings.env.php.  The con to this approach is the proliferation of settings files which serve the purpose of having different values depending on the environment.
The other solution is to just move all the environment specific settings to the pantheon preferred path, so that we have one file that serves this functionality.  The con is that it isn't exactly the drupal default - but Pantheon does not support the drupal default.
Since I don't work daily in this codebase, I'll defer to whichever you prefer.

aekong: I think we should reserve settings.local.php for local dev settings.  Pantheon has a guide on environment specific configurations:

> Use the PANTHEON_ENVIRONMENT constant and Drupal 8's override system within settings.php to enforce performance configurations based on the current Pantheon server environment.

https://pantheon.io/docs/environment-specific-config-d8

But if we want to avoid putting it in code, let's go the `settings.env.php` route and sftp it to the appropriate pantheon envs

thanks!

PR #367, opened 2019-08-26
Sg 703 search

aekong:
move search css to theme sass, pretty much just a 1-1 move


aekong:
did some more refactoring - removed an extraneous template file and updated css to reflect that

PR #378, opened 2019-09-19
SG-719: Fix News Card image issues on homepage vs department homepage

jacine:
This fixes the caching problem.  In short it:

- Creates a new view mode: `card_compact`.
- Creates a new Views display: `block_2` that renders with `card_compact` (no image).
- Ensures that the news Paragraph (used on the homepage) still/always uses the `block_1` instance, which renders `card` and contains the image.
- Adjusts the `node.html.twig` template to allow `classes` to be used from extending templates.
- Tweaks `node--news--card.html.twig` to stop extending the teaser template which was confusing.

See more details in the commit messages.

### DIFF
--- a/config/block.block.views_block__news_block_2.yml
+++ b/config/block.block.views_block__news_block_2.yml
@@ -8,14 +8,14 @@ dependencies:
     - views
   theme:
     - sfgovpl
-id: views_block__news_block_1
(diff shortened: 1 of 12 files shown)


jacine:
Closing this in favor of #379.

PR #379, opened 2019-09-19
SG-717: Ensure entire news card is clickable

jacine:
Note: This PR builds on work done in #378, which should be reviewed/merged first.

This is the only commit relevant to this PR: https://github.com/SFDigitalServices/sfgov/commit/2afeced1ea6809222f3b3fb8dc7ebc131af8f3d8


jacine:
FYI: I merged #378 with this because there were just too many changes in the same template and one, depends on the other. 👍

PR #381, opened 2019-09-19
Sg 578 group tx help

aekong:
- Added two new paragraphs: Email addresses and Phone numbers, each containing paragraphs email address and phone number, respectively.
- Added twig components and related css for each added paragraph
- Exported config for new paragraphs
- tabs to spaces and typo cleanup

### DIFF
--- a/config/field.field.node.transaction.field_help.yml
+++ b/config/field.field.node.transaction.field_help.yml
@@ -6,11 +6,17 @@ dependencies:
     - field.storage.node.field_help
     - node.type.transaction
     - paragraphs.paragraphs_type.email
+    - paragraphs.paragraphs_type.email_addresses
     - paragraphs.paragraphs_type.help
     - paragraphs.paragraphs_type.in_person_location
     - paragraphs.paragraphs_type.phone
+    - paragraphs.paragraphs_type.phone_numbers
   module:
     - entity_reference_revisions
+    - tmgmt_content
+third_party_settings:
+  tmgmt_content:
+    excluded: false
 id: node.transaction.field_help
 field_name: field_help
 entity_type: node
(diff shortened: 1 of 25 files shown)


config/field.field.node.transaction.field_help.yml
```diff
@@ -30,98 +36,109 @@ settings:
       help: help
       in_person_location: in_person_location
       phone: phone
+      email_addresses: email_addresses
+      phone_numbers: phone_numbers
```
jacine: The ordering of this in the UI, is likely something that will be complained about. Not sure if it's worth messing with now, b/c I know the others will eventually be gone.

Screen Shot 2019-09-30 at 5 01 26 PM

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--email--default.html.twig
```diff
@@ -0,0 +1,58 @@
+{#
```
jacine: There is no need to use `--default.html.twig`, `paragraph--email.html.twig` should be ok?

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--phone--default.html.twig
```diff
@@ -51,16 +51,14 @@ not paragraph.isPublished() ? 'paragraph--unpublished',
 %}
```
jacine: There is no need to use `--default.html.twig`, `paragraph--phone.html.twig` should be ok.

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--phone--default.html.twig
```diff
@@ -51,16 +51,14 @@ not paragraph.isPublished() ? 'paragraph--unpublished',
...
-            } %}
-
-        {% endblock %}
-    </div>
+  <div {{ attributes.addClass(classes) }}>
+    {% block content %}
+      {% include "@theme/phone.twig" with {
+        "phone_owner": content.field_owner.0,
+        "phone_number": content.field_tel.0,
+        "phone_details": content.field_text.0
```
jacine: I think it's better to let these render as opposed to drill to the zero, e.g. `content.field_whatever`.  For these, it's a little more work, but overriding the field templates, and just removing the wrappers is better.  You can do them individually or get all fields on x paragraph, like: `field--paragraph--phone-numbers.html.twig`

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--email--default.html.twig
```diff
@@ -0,0 +1,58 @@
...
+    view_mode ? 'paragraph--view-mode--' ~ view_mode|clean_class,
+    not paragraph.isPublished() ? 'paragraph--unpublished'
+  ]
+%}
+{% block paragraph %}
+  <div{{ attributes.addClass(classes) }}>
+    {% block content %}
+      {% include "@theme/email.twig" with {
+          "email_title": content.field_title.0,
+          "email_address": content.field_email.0
```
jacine: I think it's better to let these render as opposed to drill to the zero, e.g. content.field_whatever. For these, it's a little more work, but overriding the field templates, and just removing the wrappers is better. You can do them individually or get all fields on x paragraph, like: `field--paragraph--email.html.twig`

config/field.field.node.transaction.field_help.yml
aekong: yup, will remove them in a future config export, after data's been migrated to the new grouped paragraph fields.  thx for the reminder! 😃

aekong:
thanks for the tips!

PR #382, opened 2019-09-19
Sg 731 ci gi integration

aekong:
- Removed command to run behat tests in `.circleci/config.yml`
- Added script to trigger ghost inspector test suite for pantheon env via ghost inspector api


.circleci/scripts/ghost-inspector/run-gi-suite
```diff
@@ -0,0 +1,44 @@
...
+
+EXECUTE_URL="https://api.ghostinspector.com/v1/suites/$GI_SUITE/execute?apiKey=$GI_API_KEY&startUrl=$GI_START_URL&immediate=1"
+
+echo "Execute url: $EXECUTE_URL"
+
+RESULT_ID=$(curl -s $EXECUTE_URL | ./jq -r '.data._id')
+
+# Poll for the suite result, sleep for a few seconds if it hasn't changed
+echo "Polling for suite results (ID: $RESULT_ID)"
+while [ "$STATUS" = 'null' ]; do
```
josh-chou: is this loop necessary?
seems like STATUS will get set to empty string if the curl errors out or returns something other than json which has data.passing: true.

aekong: the suite gets triggered and can run for an unknown amount of time, so the loop polls for suite status from ghost inspector api.  but you're right, curl can error or the command `STATUS` can always be null if the api returned json that didn't have `data.passing` and the loop would never break.  i put in a check for valid json and an error response from the api so that either of those failures should break the loop.

PR #389, opened 2019-10-23
Sg 778 step by step theme

aekong:
step by step content type + theme


web/themes/custom/sfgovpl/src/sass/drupal.scss
```diff
@@ -84,4 +88,6 @@
 @import 'search';
 @import 'translate';
 
-@import 'shame';
+@import 'components.scss';
```
zakiya: Is there a more specific name that can be used? If not, maybe explain the uses at the top of the file in a comment?

PR #390, opened 2019-10-25
Sg 681 transaction icons

zakiya:


web/themes/custom/sfgovpl/dist/css/drupal.css
```diff
@@ -1732,6 +1732,40 @@ header[role="banner"] .sfgov-search-311-block form [data-drupal-selector="edit-a
   background: #a9d6ea;
 }
 
+.paragraph[data-contact] {
+  background-position: left 5px;
+  padding-left: 38px;
+}
+
+[data-contact="document"] {
```
aekong: just curious - is there a reason for using attributes instead of class names?

web/themes/custom/sfgovpl/src/sass/components/_contact-items.scss
```diff
@@ -0,0 +1,38 @@
+.paragraph[data-contact] {
+  background-position: left 5px;
+  padding-left: 38px;
+}
+
+[data-contact="document"] {
```
aekong: just curious - is there a reason for using attributes instead of class names?

PR #410, opened 2019-11-14
drupal issue 2955321 Unable to change non-translatable field value on…

aekong:
… translatable content


aekong:
will a ghost inspector e2e test suffice? unit tests and kernal tests in drupal require a lot of overhead to mock properly, enable and install modules, etc.  it's a whole separate task for all that setup!

josh-chou:
👍 as long as theres some kind of automated test coverage

PR #411, opened 2019-11-14
SG-759 - add subdomain whitelist / 404 redirect

c3gabe:


aekong:
this will be its own release.  will merge this after next release of ui bug fixes

PR #431, opened 2019-12-17
added newline

josh-chou:
updated circleci environment var to point to different location for sso certs


aekong:
Had an in-person in-depth review of this with @josh-chou

PR #437, opened 2020-01-22
Sg 886 topics 2

aekong:


web/themes/custom/sfgovpl/src/sass/node/_node-transaction-card.scss
```diff
@@ -1,32 +1,32 @@
...
-    margin: 0;
-    color: $c-dark-blue;
-    text-decoration: none;
-    &:empty {
-      display: none;
-    }
-  }
-  &:hover {
-    border-color: $c-dark-blue;
+// .sfgov-transaction-card {
```
josh-chou: Delete instead of commenting out?

aekong: removed! thanks for catching that!

PR #442, opened 2020-02-11
sg-952 - Update see more topics styling

c3gabe:


robdecker:
This looks like it is now out of date. Looking at SG-952 & the git log, I see other commits with the 952 prefix as part of the [SG-968 PR](https://github.com/SFDigitalServices/sfgov/pull/456).

@c3gabe please confirm and close this PR if it's no longer needed.

PR #443, opened 2020-02-19
sg-985 - Timeline paragraph config, styling

c3gabe:


c3gabe:
@josh-chou we've hit the multidev limit

josh-chou:
> @josh-chou we've hit the multidev limit

i deleted all the ci multidevs around the time you posted this.
see if you can join the proj-sfdotgov-eng slack channel.  tried to add you, but apparently i an not an admin.

PR #466, opened 2020-03-12
Fix news item overflow with image present, use two columns if possibl…

josh-chou:
…e (#461)

* use min-height for .news--card.has-image

* replace literal #e7b22e with $c-yellow-4

* adapt news-cards molecule to two-column layout


josh-chou:
pulling out my admin card to merge this pr.

PR #479, opened 2020-03-25
Get form.io render options from template output

shawnbot:
@hshaosf We'd like to be able to pass options to `Formio.createForm()` at runtime, on a per-form basis. (Stuff like being able to show or hide certain buttons in the [wizard](https://formio.github.io/formio.js/app/examples/wizard.html) nav, for instance.) This is an attempt to stub out the template changes that would:

1. Output the `data-options="..."` attribute as a JSON-encoded string, then
2. Parse the value as JSON and pass that to the `Formio.createForm()` function

I realize that if we encode as JSON on the template side we might have a double encoding issue — unless there's some snazzy way for Drupal to expose a JSON editor that outputs template data in object form. But in lieu of that, maybe we just take a JSON string as input (`{"foo":"bar"}`) and skip the `json_encode` filter?

### DIFF
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -49,7 +49,7 @@
 <div id="formio"
   data-source="{{ paragraph.field_formio_data_source.value }}"
   {% if paragraph.field_formio_render_options.value %}
-    data-options="{{ paragraph.field_formio_render_options.value|json_encode|escape }}"
+    data-options="{{ paragraph.field_formio_render_options.value|escape }}"
   {% endif %}
 ></div>
 <script src="https://unpkg.com/formiojs@{{ formio_js_version }}/dist/formio.full.min.js"></script>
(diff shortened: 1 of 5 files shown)


web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -46,12 +46,28 @@
 {% if paragraph.field_formio_sfds_version.value %}
   {% set formio_sfds_version = paragraph.field_formio_sfds_version.value %}
 {% endif %}
-<div id="formio" data-source="{{ paragraph.field_formio_data_source.value }}"></div>
+<div id="formio"
+  data-source="{{ paragraph.field_formio_data_source.value }}"
+  {% if paragraph.field_formio_render_options.value %}
```
hshaosf: @shawnbot  do we expect a field in the form for `render-options` similar to the `version`s fields>

shawnbot: Yes, sorry for not being clearer! That's what we'd need to support this, and I'd love your help getting that set up 😁

PR #534, opened 2020-05-14
Sg 1054 - Form.io block and Feedback form

jacine:
## Summary of changes

- Implements a new Feedback form that appears at the bottom of every page.
- Adds a custom block entity named "Form" which utilizes the Form.io paragraph (and can support other types in the future as needed).
- Adds a custom module for Form.io to store library definitions and customizations, and documentation for how to use within Drupal.
- Tweaks the existing template implementation to support multiple forms per page, and Twig blocks (so we have less code repetition where customizations are required).
- Provides a fixed content block instance, configured to print in the `content_bottom` region. This block loads on every page.  
- Removes the ability to define a version per Form.io form (see additional notes for more details).
  Screenshot of Form.io Paragraph form
- Adds a configuration page that allows changing the source of the Form.io and SDFS scripts.
    Screenshot of admin config

<details>
<summary>Additional Notes</summary>

1. I had to remove the fields and code that allows defining versioning per-form, because the second two forms are on a page, which with this PR, will be the case for every [form page](https://sfgov.gov/workers-families-first-preapproval-form), you risk loading multiple versions of both Form.io and [formio-sfds](https://github.com/SFDigitalServices/formio-sfds). I have replaced this with a global configuration. See [this comment](https://github.com/SFDigitalServices/sfgov/pull/534#discussion_r434003884) for details.
2. Despite not needing [formio-sfds](https://github.com/SFDigitalServices/formio-sfds) for the Feedback form, I had to load it anyway, because like above, if it's loaded on the page, for another form, it will be initialized, which changes a lot about the form's HTML structure and necessary CSS. Also, I realized that the SFDS theme is required for i18n functionality to work.
3. I was unable to get the forms to work using `async`.  If anyone has ideas, lmk!
4. I attempted to use a locally-hosted version of the form config, but could not figure out how to  "hook" it back up to Form.io.  <del>I left the implementation code in here, which is harmless, in case anyone wishes to try it out.  Given an endpoint of `https://sfds.form.io/workers-families-first` it will look for `web/modules/custom/sfgov_formio/data/workers-families-first.json`, and if found, it will pass the `config` as JSON directly to `Formio.createForm`</del>.

</details>

## Testing Notes

- Visit https://pr-534-sfgov.pantheonsite.io/. The Feedback form should be at the bottom of the each page.
- Visit https://pr-534-sfgov.pantheonsite.io/workers-families-first-preapproval-form, which is an example of a page that has two instances on a page.
- Visit https://pr-534-sfgov.pantheonsite.io/node/973, which is the building permit form (currently under development)

## Deployment Notes

Configure the fixed content block's content:

1. Load the site (e.g. homepage). This creates the fixed block content instance.
1. Log in, and go to `/block/2`.
2. Click "Add Form.io"
3. In "Data Source" field, add: `https://sfds-test.form.io/feedback `
4. In "Form.io Render Options" field, add: 

    ```json
    {
      "formioSFDSOptOut": true,
      "i18n": "https://i18n-microservice-js.herokuapp.com/google/1ialzMAruk1wk2sIAR-WHzlCsJ73yy7LIxruD3cvTYFE"
    }
    ```
5. Save.

You should now see the form at the bottom of each page.

feedback-form

## Todo

- [x] Get help testing existing forms, in the hopes of removing the version fields entirely (they are not functional with this PR anyway).
- [x] Get proper access to Airtable and Form.io to move the form and data outside of my personal account ;)


web/modules/custom/sfgov_formio/README.md
```diff
@@ -0,0 +1,8 @@
+# SF Gov Form.io
+
+This module stores the library definitions, and related code for Form.io forms.
+
+## Todo
+
+- [ ] Remove the ability to choose a version for Form.io and Form.io SFDS packages from the UI. This requires testing existing forms first.
+- [ ] Complete and document locally hosted config option, or remove it.
```
jacine, line 8: @shawnbot @hshaosf @aekong Hey y'all.  Could you give this a quick review when you get a chance? I wrote up some details in the Notes section of the PR description.  I need to resolve these two issues before merging, which Nicole would like to do.

hshaosf: @jacine thanks for all the great work you are doing!  I'm curious does the documentation include how do we configure a sandbox to use specific versions of the packages?  Thinking of what our workflow would be like to review changes for new updates.  Happy to hear other folks' thoughts on this as well.

jacine: Thanks for taking a look at this @hshaosf!  I can definitely write that up. Good idea!   In the meantime, this is what you would do:

1.  You change the URL in [`sfgov_formio.libraries.yml`](https://github.com/SFDigitalServices/sfgov/pull/534/files#diff-49c223d56d245109546d7ca82e3ff4cdR9) to point to whatever version you need.
2. Clear the cache.
3. If you need a sandbox environment, commit, push and open a PR (so you can get a multidev env)

Do you think you need a way to do this without touching the codebase?  I could probably create a settings page for the module that would allow you to alter the library URL in the admin interface for testing (sort of like a debug setting). I think this depends on how y'all work over in Form.io. I know rapid changes are happening, and I worry that NOT doing this could slow things down/make it a poor DX.  But, I also know we can't support 2 versions at the same time right now, so we need a happy middle ground.  What do you think?

I'll get the docs in the module README shortly. ;)

aekong: @jacine you are 🔥 ! i'm going to pull this down and go through the steps today.  i haven't gotten into form.io land yet, so this will be helpful.

as far as a settings page to alter the library url, i think that'd be the best route.  our current workflow requires (usually) a pr, review, build, test, deploy which will probably delay the rapid iterations for @hui and @shawnbot and company.  

Buuut I'm also not sure how to get around that 2 versions thing.  I'm actually not sure what the situation is there, so I may have different thoughts after pulling this down.

jacine: Thanks @aekong! I appreciate it. I'll get started on a debug config setting.

> Buuut I'm also not sure how to get around that 2 versions thing. I'm actually not sure what the situation is there, so I may have different thoughts after pulling this down.

Yeah, so I can help clarify this quickly for you.  The current implementation has fields that allow the person adding the form to specify which version of both Form.io AND the SFDS theme they want to use.  It looks like this, which you can see [here](https://pr-534-sfgov.pantheonsite.io/node/900/edit) if you login via `terminus drush sfgov.pr-534 uli`:

Screen Shot 2020-05-29 at 12 08 48 PM
 
It then uses that input to load the JavaScript for these:

https://github.com/SFDigitalServices/sfgov/blob/47377ad772e077b281d26471d8294d817d1ed801/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig#L55-L56

This was okay, until this PR, as the only place Form.io forms were used were in the "Form page" content type.  There would only be one per page, and no chance of conflicts or other issues.

This PR adds a Feedback form that loads on ALL pages.  Originally, when I worked on this, I tried to keep the implementation as lean as possible.  I didn't need the Form.io SFDF theme, so I didn't use it at first.  But then, I tried testing a Form page [example](https://pr-534-sfgov.pantheonsite.io/workers-families-first-preapproval-form)... It uses the Form.io SFDS theme and basically everything in my form broke.  The SFDS theme _really_ customizes a lot (completely different markup, classes and CSS), and it automatically initializes on any forms it finds.

If we leave the original implementation, we risk loading 2 different versions of each, on pages where there are multiple forms, and all sorts of pain. So these version fields have to go.  I didn't actually remove them yet, because I needed feedback from y'all, but they don't actually DO anything in this PR, which removes the lines of code I referenced above and loads the scripts via a Drupal library.  ;)

shawnbot: One thing that we could do is add the ability to opt _out_ of the theme styling in the Formio options, for instance (making up

```js
Formio.createForm(feedbackFormElement, {
  formioSFDSoptOut: true,
  ...otherOptions
})
```

And in a new release of the theme JS we would detect that option and _not_ add the `formio-sfds` classname that scopes all of the styles.

I'm usually not a fan of this type of thing, but it seems like a low-risk and clear way to do it. Thoughts, concerns?

jacine: That's definitely an option, and a good one for the Feedback form, since it would help avoid load the SFDS theme JS on all pages. I'm all for it. 👍 

It wouldn't solve the possibility of multiple conflicting versions, so we'll still need to deal with that.  Working on a debug configuration page for that, so you all can test newer versions more easily.

jacine: Hi y'all!

I just pushed an update here.  Highlights:

## 1. Removed Version Fields from Paragraphs

I also added some descriptions to the form fields.Here is a screenshot:

formio-paragraph

## 2. Added Admin Configuration Page

This page allows you to override the source location of both library definitions.  You can try it out here: https://pr-534-sfgov.pantheonsite.io/admin/config/services/sfgov_formio. It looks like this:

formio-config

## 3. Added Documentation

I added details about the implementation, how to add Form.io content, the library definitions, and how to override them in the admin UI.  View the README here: https://github.com/SFDigitalServices/sfgov/blob/452f7f09ecfbcf3e7a97eeb67bcc061745f3116e/web/modules/custom/sfgov_formio/README.md

P.S. We can implement `formioSFDSoptOut` as a follow up whenever that is ready. ;) 

Lmk your thoughts!
Jacine :)

shawnbot: @jacine You can test the `formioSFDSOptOut` option if you use the `0.0.0-3166e1f` canary release from https://github.com/SFDigitalServices/formio-sfds/pull/42. Let me know how it goes!

jacine: Hey thank you @shawnbot!  I switched the script out here, so we are using:

```html
<script src="https://unpkg.com/formiojs/dist/formio.full.min.js"></script>
<script src="https://unpkg.com/formio-sfds@v0.0.0-3166e1f/dist/formio-sfds.standalone.js"></script>
```

And I've added the opt out option to the Feedback form. I'm getting the `SFDF form opted out` console message, but all the markup customizations still seem to be applied.  Is that expected?  I expected the "Workers families first" form on [this page](https://pr-534-sfgov.pantheonsite.io/workers-families-first-preapproval-form) to remain as is, and the Feedback form styles to break.

jacine: Updated this again to capture changes in #561.

Note: I forgot that I can't use formioSFDSOptOut, if I want to use translations, which I do!  So, the setting is no longer configured in the PR instance, and this should be ready to go.

jacine: @shawnbot I think I managed to mess up the options before. Just wanted to let you know [this is working now](https://pr-534-sfgov.pantheonsite.io/), just fine with the opt out setting AND translations!  I also tested with `6.0.0-rc.fb5b979` and that also works great (and I absolutely needed this `formioSFDSOptOut` option, because without it the custom events were overriding by `submitdone`)!  So yay, and thank you!!  🎉

jacine:
> @jacine I'd like to publish the 6.0 release of the theme ([SFDigitalServices/formio-sfds#49](https://github.com/SFDigitalServices/formio-sfds/pull/49)) before you deploy this. I wouldn't consider it a blocker, but having both formiojs and formio-sfds pinned to release candidates is not ideal. It looks like there are some ugly merge conflicts to resolve, so maybe I have some time? 😬

Hey @shawnbot!  Oh, no it's not pinned! I think I confused you... This PR allows you to change the settings in the admin UI, but by default it's still using `latest` without the query lookup as usual.  
```yml
formio:
  js:
    'https://unpkg.com/formiojs/dist/formio.full.min.js': { type: external }

formio_sfds:
  js:
    'https://unpkg.com/formio-sfds/dist/formio-sfds.standalone.js': { type: external }
  dependencies:
    - sfgov_formio/formio
```

This requires testing any forms that were pinned to older or other versions (which I believe happened already), but it's going to use this ⬆️ , unless overridden in the UI (which doesn't happen here, and should only happen on dev for testing).  It's basically a quick debug setting for y'all so you don't have to edit code to test your forms on new versions: 

What I was trying to say above is that I _tested_ it with the 6.0 release, in addition to the current, and all is well.  The Feedback form PR doesn't need the release, but I did need the opt out setting, because without it 6.0 would have broken my form. That was all.  Hopefully that clarifies!  LMK if I missed something.

jacine:
And I did not realize there were conflicts here! Thank you!! 👍

PR #543, opened 2020-05-27
center background img for spotlight

aekong:


aekong:
reviewed by @c3gabe locally.

PR #560, opened 2020-06-08
Confirmation page redirects

henryjiang-sfgov:
Add custom logic to allow custom redirect after form submission.


henryjiang-sfgov:
Getting too messy, close this out.

PR #561, opened 2020-06-08
Add redirect logic for form.io forms

henryjiang-sfgov:
**1. Update confirmation page template per [Trello task ](https://trello.com/c/pQoODGZH/13-henry-implementation-for-sfgov-confirmation-page)
2. Add custom logic to allow custom redirect after form submission**

**Files changed**

- paragraph--form-io.html.twig  -- Update form-io paragraph template.  The javascript code works off the Form render options, new property "redirects" has been added.
- node--form-confirmation-page--full.html.twig -- Update confirmation page template
- hero-banner-with-image.twig  -- New component for banner with image
- _node-form-confirmation.scss -- CSS source for form confirmation pages
- drupal.css -- compiled CSS

The rest of the files are configs


web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -62,6 +62,16 @@
       .then(function(form) {
       // What to do when the submit begins.
       form.on('submitDone', function(submission) {
+          // custom options defined in Form.io render options field
+          if(options['redirects'] !== undefined){
```
shawnbot, line 66: One problem with testing for equality with `undefined` is that `null !== undefined`. I'd suggest either:

```suggestion
          if (options.redirects instanceof Object) {
```

or simply:

```suggestion
          if (options.redirects) {
```

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -62,6 +62,16 @@
...
+            for (var key in options['redirects']) {
+              var value = options['redirects'][key]
```
shawnbot, line 68: I'd also go with the dot syntax here instead of brackets:

```suggestion
            for (var key in options.redirects) {
              var value = options.redirects[key]
```

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -62,6 +62,16 @@
...
+               // only one "redirect" should be "yes", this is set in the form.io form
+              if(submission['data'][key] !== undefined && submission['data'][key] === "yes")
```
shawnbot, line 70: This condition is redundant: If `submission.data[key] === 'yes'`, then it will not be `undefined`.

```suggestion
              if (submission.data[key] === "yes")
```

henryjiang-sfgov: If key doesn't exist in submission data, wouldn't this throw an exception?

henryjiang-sfgov: @shawnbot please review my updated code. Thanks.

shawnbot: > If key doesn't exist in submission data, wouldn't this throw an exception?

Nope! `({data: {}}).data.foo` doesn't throw an exception, but `({}).data.foo` does (because `.data` is undefined).

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -62,6 +62,16 @@
       .then(function(form) {
       // What to do when the submit begins.
       form.on('submitDone', function(submission) {
+          // custom options defined in Form.io render options field
+          if(options.redirects instanceof Object){
+            for (var key in options.redirects) {
+              var value = options.redirects[key]
+               // only one "redirect" should be "yes", this is set in the form.io form
+              if(submission.data[key] && submission.data[key] === "yes")
```
shawnbot, line 70:
```suggestion
              if(submission.data[key] === "yes")
```

PR #581, opened 2020-06-19
sg-1119 - add notranslate class to content if drupal translation exists

aekong:


web/themes/custom/sfgovpl/src/js/translate.js
```diff
@@ -107,6 +107,9 @@ function SFGovTranslate() {
       // If translated content is being shown on the page, it should be wrapped
       // in a container with class="notranslate" to allow other elements like
       // header and footer to be translated.
+      if(that.getDrupalTranslation(currentDrupalLanguage)) {
+        $('main[role="main"]').addClass('notranslate');
```
shawnbot, line 111: We may also want to add [`translate="no"`](https://www.w3.org/International/questions/qa-translate-flag.en) for other services:

```suggestion
        $('main[role="main"]')
          .addClass('notranslate')
          .attr('translate', 'no');
        
```

aekong: yessir

PR #586, opened 2020-06-29
Fix config export issues

jacine:
This doesn't make any actual changes.  Current config doesn't ever report as imported due to sorting issues, which this resolves.

### DIFF
--- a/config/core.entity_form_display.node.meeting.default.yml
+++ b/config/core.entity_form_display.node.meeting.default.yml
@@ -179,9 +179,9 @@ content:
       form_display_mode: default
       default_paragraph_type: _none
       features:
-        duplicate: duplicate
-        collapse_edit_all: collapse_edit_all
         add_above: '0'
+        collapse_edit_all: collapse_edit_all
+        duplicate: duplicate
     third_party_settings: {  }
     region: content
   field_meeting_cancel:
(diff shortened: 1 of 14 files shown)


aekong:
nice! so doing `drush cim -y` imports the config, then `drush cim -y` will now report that there are no new configurations to import, where previously it would just import the config again? i was wondering why that was happening.  did you have to manually sort these configs? how did you go about discovering that sorting was the issue? sorry for all the questions, just trying to avoid this in the future.  thanks!

jacine:
> nice! so doing `drush cim -y` imports the config, then `drush cim -y` will now report that there are no new configurations to import, where previously it would just import the config again? 

Exactly!

> i was wondering why that was happening.

It's usually due to Git merging issues. Either that or hand-editing config.

> did you have to manually sort these configs? how did you go about discovering that sorting was the issue? sorry for all the questions, just trying to avoid this in the future. thanks!

Nope. I was getting setup to work on a new task, so doing the usual env reset locally, and I imported config, and noticed the same stuff pending from last week.  So, I tried `lando drush cim -y` again, and noticed it was still reporting the same files.  At that point, I just ran `lando drush cex -y`, and I could see from the diffs it was just random sorting issues.  Then, I confirmed with another import that it was resolved (no changes pending), so I committed and pushed.

🎉

aekong:
🎉

PR #611, opened 2020-07-20
Pin formiojs and theme versions

shawnbot:
This pins the versions of formiojs and the SFDS theme to specific versions so that the settings persist in other environments (test-sfgov and PR deployments). @jacine and I discussed this in Slack, and my thinking is that it's better to have these in git rather than in the Drupal UI.

As for the specific versions, I think it's important to note that:

1. We're "stuck" on formiojs v4.10.4 until I can identify whatever is causing the bizarre issue with date/time input icons rendering as `[Object HTMLElement]` (citation needed!).
2. I've upgraded us to the [6.0 release of formio-sfds](https://github.com/SFDigitalServices/formio-sfds/releases/tag/v6.0.0), which gives us Truss's design updates and more accessible form navigation.


aekong:
this should also fix [sg-1197](https://sfgovdt.jira.com/browse/SG-1197)

PR #612, opened 2020-07-20
SG-1095: form confirmation page css

henryjiang-sfgov:
Update form confirmation page CSS

### DIFF
--- a/web/themes/custom/sfgovpl/templates/components/hero-banner-with-image.twig
+++ b/web/themes/custom/sfgovpl/templates/components/hero-banner-with-image.twig
@@ -8,7 +8,7 @@
      {% endif %}
     {% if banner.text %}
     <div class="banner-text">
-      <H1>{{ banner.text }}</H1>
+      <h1>{{ banner.text }}</h1>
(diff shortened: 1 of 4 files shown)


web/themes/custom/sfgovpl/templates/components/hero-banner-with-image.twig
```diff
@@ -8,7 +8,7 @@
      {% endif %}
     {% if banner.text %}
     <div class="banner-text">
-      <p>{{ banner.text }}</p>
+      <H1>{{ banner.text }}</H1>
```
aekong, line 11: nitpick: can this be lowercase `h1`?
```suggestion
      <h1>{{ banner.text }}</h1>
```

henryjiang-sfgov: No this pr didn't make any CSS changes, apparently it didn't need to.  Changes were made in the node field template level, that probably updated drupal.css? I supposed we can pull drupal.css and see how it looks.

PR #632, opened 2020-08-13
[SG-1153] Fix conflict-related issues.

jacine:
Caused by #592 and #606 working on the same filters.


web/themes/custom/sfgovpl/sfgovpl.info.yml
```diff
@@ -20,7 +20,6 @@ libraries:
   - sfgovpl/sfgov-components
   - sfgovpl/sfgov-back-to-top
   - sfgovpl/sfgov-in-this-page
-  - sfgovpl/sfgov-campaign
```
jacine, line 23: @aekong I let this one slip back in accidentally i think, after your awesome CSS cleanup.  It's being loaded in the template.

aekong: thanks for fixing so quick!

PR #634, opened 2020-08-18
sg-1203 - put back to top link in twig so amplitude js registers the …

aekong:
Amplitude js initializes tracking events before this link gets inserted into the dom.  This pr moves the markup for the back to top link from being created dynamically via javascript to the `html.html.twig` template.

### DIFF
--- a/web/themes/custom/sfgovpl/src/js/back-to-top.js
+++ b/web/themes/custom/sfgovpl/src/js/back-to-top.js
@@ -8,8 +8,6 @@
 (function ($, Drupal) {
   Drupal.behaviors.backToTop = {
     attach: function (context) {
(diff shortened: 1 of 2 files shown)


aekong:
@minnur let me know what you think about the approach here.  not sure why the "back to top" link was previously being inserted into the dom via js, but since amplitude initialization happens before the link is added, the link doesn't get registered for event tracking.

@nicolesque if this change is accepted, then do you still want to track this "back to top" click explicitly? With this change, the "back to top" link is properly tracked as a link click event.  If we want log this click explicitly as `event_type` **Back to top** we can leave the configuration changes in.  Otherwise we can remove the config and it will get logged as `event_type` **Link click**.

Here's what will get sent to amplitude with the additional configuration:

image

PR #644, opened 2020-08-26
Sg 1205 Accessibility/bug fixes

jacine:


web/modules/custom/sfgov_utilities/sfgov_utilities.module
```diff
@@ -250,4 +251,31 @@ function _sfgov_utilities_build_map($address) {
...
+  }
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function sfgov_utilities_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Attach #form_id to all form elements.
+  sfgov_utilities_attach_form_id($form, $form_id);
+}
```
arlina-espinoza, line 281: I'm curious as to why is this needed?

jacine: We need to make sure we are only targeting the checkboxes wrapper in the meeting filters form, and elements don't have that context, so we are passing that down to elements in here. The preprocess hook closer to the top is using this further up in the diff. 😀

PR #657, opened 2020-09-08
Use GitHub deployments API

shawnbot:
This adds a step to our `build_deploy_and_test` CircleCI job that posts a deployment + status to the [GitHub deployments API](https://developer.github.com/v3/repos/deployments/#create-a-deployment). This doesn't replace @josh-chou's multidev link comments, but instead adds deployment events to the timeline of each PR that look like this:

image

You can also find them in the [repo deployments view](https://github.com/SFDigitalServices/sfgov/deployments):

image

This also simplifies setting up a GitHub Actions workflow for Lighthouse CI that can run in parallel with the CircleCI builds (without blocking them!), and which adds Lighthouse score statuses to every push. You can see what this looks like for DAHLIA over in https://github.com/SFDigitalServices/sf-dahlia-web/pull/1411.

Another thing I'd love to do in a future PR is change the `GITHUB_TOKEN` secret in CircleCI to a GitHub "bot" account so that @josh-chou doesn't personally have to do all of this work for us. 😉


.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ defaults: &defaults
     # See: https://discuss.circleci.com/t/circle-2-0-global-environment-variables/8681
     #======================================================================================================================================
 
-    NOTIFY: 'scripts/github/add-commit-comment {project} {sha} "Created multidev environment [{site}#{env}]({dashboard-url})." {site-url}'
+    NOTIFY: 'scripts/github/create-deployment {project} {sha} {site-url}'
```
shawnbot, line 34: @josh-chou I couldn't find any documentation for this `NOTIFY` field in the CircleCI config. I assumed that this is how the commit comments linking to the multidev deployment were created, but the `create-deployment` script I've created in this PR doesn't appear to be running in this PR. Any tips?

josh-chou: Our drupal was set up with https://github.com/pantheon-systems/example-drops-8-composer.
That's where this file came from.

josh-chou (dismissed):
lgtm

.circleci/config.yml
```diff
@@ -30,8 +30,6 @@ defaults: &defaults
     # See: https://discuss.circleci.com/t/environment-variable-expansion-in-working-directory/11322
     # See: https://discuss.circleci.com/t/circle-2-0-global-environment-variables/8681
     #======================================================================================================================================
-
-    NOTIFY: 'scripts/github/add-commit-comment {project} {sha} "Created multidev environment [{site}#{env}]({dashboard-url})." {site-url}'
```
shawnbot, line 34: I had to dig deep into some Pantheon build tools to figure out where the deployment comments were _actually_ coming from, and it turns out: not this `NOTIFY` environment variable! Even though [that line still exists in the template repo](https://github.com/pantheon-systems/example-drops-8-composer/blob/e4905fab55958a1da8ff202f76b8478af8c2f0e5/.circleci/config.yml#L51), [the call to `terminus build:env:create`](https://github.com/pantheon-systems/example-drops-8-composer/blob/e4905fab55958a1da8ff202f76b8478af8c2f0e5/.ci/deploy/pantheon/dev-multidev#L20) doesn't use it, and the `notify` option [is deprecated and ignored](https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/d0ad92df58fb3c89626a7d79a719f53118976a2a/src/Commands/EnvCreateCommand.php#L29) anyway.

TL;DR: this is unused. The comment contents are [hard-coded into the terminus `env:create` command](https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/d0ad92df58fb3c89626a7d79a719f53118976a2a/src/Commands/EnvCreateCommand.php#L154).

scripts/gitlab/add-commit-comment
```diff
@@ -1,17 +0,0 @@
-#!/bin/bash
```
jacine, line 1: This is fine to delete (this whole folder actually), but you meant to delete from the github folder, not gitlab.

scripts/github/update-deploy-status
```diff
@@ -0,0 +1,57 @@
...
+env="$TERMINUS_ENV"
+transient='true'
+auto_inactive='true'
+
+if [ "$CIRCLE_BRANCH" = "main" ]; then
+  env='production'
+  transient='false'
+  auto_inactive='false'
+  target_url="https://sf.gov"
+fi
```
jacine, line 18: I am not sure, but I _think_ the `env` still be `$TERMINUS_ENV` and `target_url` should be `DEV_SITE_URL`?  Also, not sure this should be `production`. We might have to see how this goes when we merge!

Technically merges to `main` go to Pantheon `dev` and then there is no connection to this Github repo anymore. Once at dev, deployments need to be pushed through Pantheon's dashboard to `test` and then `prod`.

jacine (commented):
You are awesome @shawnbot. That is all! 👏  🥇 🧙 

Posted a few comments. 😉

scripts/github/update-deploy-status
aekong: yeah you might need to hook into patheon's deploy hooks to get that info to gh actions.  those are in `pantheon.yml`

jacine: @aekong I think @shawnbot has the right variables generally (this is working perfectly for this PR, and so would be fine for multidev envs)... I just don't _think_ the merge to `main` is actually special, other than not being transient.

scripts/gitlab/add-commit-comment
shawnbot: I actually did mean to delete this, since we don't use gitlab. But if this makes updating our deployment scripts harder in the future, then I'm happy to revert the commit.

scripts/github/update-deploy-status
shawnbot: Interesting! Yeah, I was definitely glossing over how this would work for the production deployment.

In this PR, I'm really just using the deployments API as a more formal notification layer than commit statuses. Ideally, the deployment platform itself (Pantheon) would be responsible for [posting a deployment status](https://docs.github.com/en/rest/reference/repos#create-a-deployment-status) every time the state of the deployment changes (pending, success, failure, inactive, etc.), but I'm just assuming that it's finished and creating a new deployment + success status.

But you're right, @aekong: Responding to webhooks would be the best way to handle this, and it would allow us to lean on Pantheon as the source of truth. For instance, Pantheon should be able to tell us when a multidev environment becomes inactive, and the GitHub UI would reflect the state of those environments in ~realtime.

Let's leave this unmerged for now, and I'll look into Pantheon webhooks. 👀

shawnbot: Well, [that was fast](https://pantheon.io/docs/continuous-integration#known-limitations):

> **Known Limitations**
> At this time, Pantheon does not provide or support:
> - Webhooks

Is there something we can do in `pantheon.yml` to run a script whenever a deployment status changes?

aekong: @shawnbot i think these are the only workflow hooks pantheon support right now: https://pantheon.io/docs/quicksilver

shawnbot: Okay, it looks like I'm going to have to [brush up on my PHP](https://pantheon.io/docs/quicksilver#hooks) 😱

jacine: 😂 I still think this is a great improvement.  So what If this doesn’t deal with production deploys, you know? It’s still better than the pantheon comment!

scripts/gitlab/add-commit-comment
jacine: @shawnbot Oh, sorry! Nah, no need to revert.  I just thought you meant to delete the `scripts/github/add-commit-comment`.

scripts/github/update-deploy-status
aekong: yeah i would say mission accomplished at this point! 👍 🎉

jacine (approved):
This is great!  I think the merge to main part is incorrect, but we can check that out separately in a PR where it can actually be debugged if necessary. ;)

aekong (approved):

PR #658, opened 2020-09-09
Sg 1184 people section

aekong:
* add common styles for people section
* make exception for front page people section


jacine (approved):
👍

PR #660, opened 2020-09-10
Set up Lighthouse CI

shawnbot:
Okay, so I'd love to test this out on the develop branch and see if it works, because the whole poll-the-deployments-API approach is clearly not working. Here's what _should_ happen after this merges:

Every time a successful `deployment_status` event is triggered, Actions will checkout the repo at the same git SHA/ref as the deployment and run LHCI. I've had to make [another action](/sfdigitalservices/git-the-branch) to determine the right git branch from these events, since (unlike `push` events) they're not associated with a branch/ref. But it seems to work, and now instead of having a Lighthouse action run tied to every push, they only run on deployments, and the commit still gets the status:

image

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -11,7 +11,7 @@ jobs:
         uses: SFDigitalServices/wait-for-deployment-action@v2
         with:
           github-token: ${{ github.token }}
-          interval: 300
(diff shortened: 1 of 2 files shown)


jacine (dismissed):
🤞

aekong (approved):

PR #663, opened 2020-09-11
Quicksilver → GitHub deployments API

shawnbot:
Let's see just how rusty my PHP skillz are 😭


aekong:
hi sorry for the late response.  i don't have a great way to test this.  what i've been doing  is `lando drush scr /path/to/script` but that was just for slack notifications and such.  It looks the stuff you're doing actually needs the pantheon server vars.

It looks like the deploy hook you're targeting `create_cloud_development_environment` will execute the script at `web/private/scripts/deploy_status/update.php` in a multidev context (according to the [quicksilver docs](https://pantheon.io/docs/quicksilver) anyway).  let's try to rebuild this pr and see what happens.

web/private/scripts/deploy_status/update.php
```diff
@@ -0,0 +1,87 @@
...
+
+require '../shared.php';
+
+$workflow_type = $_POST['wf_type'];
+print("Deployment status update for workflow '$workflow_type'...");
+
+$secrets = _get_secrets(['github']);
+$token = $secrets['github'];
+$environment = $_ENV['PANTHEON_ENVIRONMENT'];
+$repo = 'sfdigitalservices/sfgov';    # FIXME: needs to be generalized
```
shawnbot, line 11: @aekong is this an environment variable or something we could get elsewhere rather than hardcoding it?

PR #666, opened 2020-09-16
[SG-1223] Fix link color in Feedback form success message.

jacine:


jacine:
This code went in #668. Closing.

PR #678, opened 2020-09-24
[SG-1138] News Card Refactor

jacine:
- Use the proper view mode when there is an image (homepage) vs not (topics, departments).
- Remove homepage-specific CSS, which as using a different layout technique.
- All the News card CSS lives in one place, and is documented.
- Change the breakpoint for when the image appears, as there is never enough room for it to look good until `$narrow-screen`.


c3gabe (approved):
Looks good

PR #692, opened 2020-10-02
Composer lock update + outstanding configuration

jacine:
This gets both composer and Drupal config in line (no pending imports). I believe the manifest was missing from #628, because composer install likely wasn't run after updating the branch or after fixing conflicts.  Either way, this gets everything into a good state.


aekong (dismissed):
Thank you for syncing these up!

aekong (approved):

PR #700, opened 2020-10-19
SG-1245: add null check before appending css class to element

henryjiang-sfgov:
There is a setting in the form page's form.io render options to add a css class to the block-gtranslate element, the purpose is align the translation icon on mobile version.

Fixed in `paragraph--form-io.html.twig`


jacine (approved):
LGTM! Thanks @henryjiang-sfgov 👍

PR #702, opened 2020-10-20
[SG-1246] Implement view for unpublished content.

minnur:


web/themes/custom/sfgovpl/templates/layout/page--403.html.twig
```diff
@@ -0,0 +1,16 @@
...
+  <div class="region-fourofour-footer">
+    <div class="access-denied-message">
+      <div class="access-denied-header">
+        <h1>This page is not available</h1>
+      </div>
+      <div class="access-denied-text">
+        <p>You can try searching or <a href="{{ url('<front>') }}">go to the home page</a>.<br />
+          If you need help, <a href="https://sf311.org/" target="_blank">call 311</a>.</p>
+      </div>
+      <p>If you are a content publisher, log in to access this page.</p>
```
arlina-espinoza, line 13: I think we should run the text through translations, as the 403 might come from a non-english page.

arlina-espinoza (changes requested):

web/themes/custom/sfgovpl/templates/layout/page--403.html.twig
minnur: Yep just added that.

aekong (approved):

aekong (approved):

aekong (approved):

aekong (approved):

PR #715, opened 2020-10-27
Fix fatal error on unpublished campaigns.

jacine:
Fixes an error currently happening on http://sf.gov/ask-about-prep (while anonymous).

### DIFF
--- a/web/themes/custom/sfgovpl/includes/page.inc
+++ b/web/themes/custom/sfgovpl/includes/page.inc
@@ -1,5 +1,7 @@
 <?php
 
+use Drupal\node\NodeInterface;
+
 function sfgovpl_preprocess_page(&$variables) {
 
 // The following tests the error messages.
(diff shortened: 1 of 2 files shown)


minnur (dismissed):
LGTM! I think !empty($node) would also solve the same issue.

web/themes/custom/sfgovpl/includes/page.inc
```diff
@@ -16,10 +16,10 @@ function sfgovpl_preprocess_page(&$variables) {
  * Implements theme_preprocess_HOOK().
  */
 function sfgovpl_preprocess_page__node__campaign(&$variables) {
-  $node = $variables['node'];
-
   // Conditionally render Logo or Title.
-  if (isset($node)) {
+  if (!empty($node)) {
```
jacine, line 20: @minnur You're right! Fixed. I thought this was necessary for node previews, but it looks like that is no longer the case... Though, previews are also broken here, and throwing lots of notices. I added a fix for those as well.

minnur (approved):
LGTM

PR #716, opened 2020-10-28
[SG-1243] Add custom module (alter address field).

minnur:

PR #717, opened 2020-10-28
[SG-1243] [WIP] In-Person to Address Entity, Prep for removal of Mailing Address

jacine:
- [x] Backend Configuration
- [x] Prevent Mailing Address from being used going forward.
- [x] Customizations to Field Widget (edit form labels, descriptions, etc)
- [x] All front-end

This branch also contains @minnur's branch from #716, which is good to have separately in case we want to commit early, which is safe to do before we are ready with everything, as his PR does not enable it. 

## Deployment Notes

1.  Need to run `drush entity-updates`.  Needed for `location.field_address`.


jacine:
Closing in favor of #727.

PR #725, opened 2020-11-04
Deploy issue image attrib

aekong:


config/core.entity_view_display.paragraph.spotlight.default.yml
```diff
@@ -39,7 +39,7 @@ content:
     weight: 4
     label: hidden
     settings:
-      view_mode: default
+      view_mode: image_without_attribution
```
jacine, line 42: Hey @aekong, I just figured out what happened here, and this PR shouldn't be necessary. 

There was likely a merge conflict issue.  The `image_attributes` "field" should only be configured to show in the "Preview" view mode.  All front-end facing versions of this are handled in their respective templates. There is also potential for variable clashing in some circumstances, for two different things.  Just posted #726, which reverts this PR, fixes the display config, and renames the `image_attributes` variable in the templates, which is something else entirely. Can you take a look when you get a chance?

PR #726, opened 2020-11-04
[SG-1152] Rename spotlight image attributes variable.

jacine:


aekong (approved):
ah i see, thanks for fixing @jacine!

PR #734, opened 2020-11-12
SG-804 - update entity_browser and patch to delay modal dialog positi…

aekong:
…oning


jacine (approved):

aekong:
Thanks for reviewing @jacine !

PR #738, opened 2020-11-16
[SG-1143] Permissions version 1 (For reference only)

arlina-espinoza:
To do:
- [ ]  /admin/content should be filterable or sortable by Ready For Review
- [ ] Do we need teaser display settings in this branch?
- [x] When departments are removed, reviewer list is not updated.
- [x] QA transaction bug thoroughly.
- [x] verify permissions
- [x] remove backend (form) validation when no reviewer is selected and content is marked ready for review. 
- [x] Add front end validation suggesting a user pick a reviewer.
- [x] Fix js error - Uncaught TypeError: Cannot read property '_triggering_element_name' of undefined from parents-departments (viewable on transaction pages)
- [x] add Digital Services role in separate PR
- [x] Refactor code in sfgov_moderation.module

Edge cases:
- [x] Perms are derived from "current" revision not "latest" revision. So if user changes the dept on a published node, chaos ensues.
- [ ] If the reviewer edits the node, it won't stay assigned to them because we removed current user from review options list.


zakiya:
This PR is moot since we changed our approach and decided not to have permissions related to departments. I'm keeping it open as a reference until PR-945 is merged.

zakiya:
This work was achieved in PR-945.

PR #739, opened 2020-11-18
[SG-1269] Diffs and Content Moderation

jacine:
1. Customizes the Content Moderation Widget
2. Customizes the Preview form widget.
3. Adds the Diff module.

## Moderation

ContentModerationWidget

## Previews

Node Previews

## Diff

NodeDiff

Note: While the "inline" screenshot shows it looking perfect, that is NOT the case on most content types. There are various reasons for this, mostly related to how things are done in the theme, but also because Diff module strips all style tags (which removes a lot of our background styles for things like campaign spotlights, and because it does not use the proper templates when rendering... I have some ideas for improving this experience, that I will bring up when there is time.

_UPDATE (1/12/21):  I think that using a separate theme for inline diffs would potentially be a good solution. Maybe Gin, maybe even Seven. Making this work visually with the front end theme is not going to be viable, as even if we put in the work required, there would still be major issues due to the wqy the Diff module works, and does not use the appropriate template suggestion when generating the rendered HTML._

### DIFF
--- a/composer-manifest.yaml
+++ b/composer-manifest.yaml
@@ -10,6 +10,7 @@ packages:
     behat/mink-goutte-driver: v1.2.1
     behat/mink-selenium2-driver: v1.4.0
     behat/transliterator: v1.3.0
+    caxy/php-htmldiff: v0.1.9
     commerceguys/addressing: v1.0.9
     composer/ca-bundle: 1.2.8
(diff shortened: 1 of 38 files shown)


zakiya (approved):
Very impressive!

PR #736, opened 2020-11-14
Add amplitude events for form.io forms

shawnbot:
This PR adds log Amplitude events for more general timing related bits:

- `window.load` with `load_time` from the start of the form paragraph `<script>` execution to window `load` event firing, in milliseconds
- `form.load` with `load_time` from the window `load` to when the `Formio.createForm()` promise resolves, also in milliseconds

And for more specific form interactions. The below events include a `form` object with, at the very least, a `url` key indicating the form.io data source URL. Additional event data noted below is nested in the event's `form` object.

- `form.submit` before the submission is sent to form.io (no additional data)
- `form.submitDone` after successful submission (no additional data)
- `form.submitError` when a submission fails validity checks (`component` is the component key, and `error` is the validity error message)
- `form.error` whenever an error occurs (`error` is the error message)
- `form.componentError` for errors specific to a component, including validity failures (`component` is the component key, and `error` is the error message)
- `form.prevPage` when the "Back" form nav button is clicked (`page_index` is the _destination_ page index, starting from zero)
- `form.nextPage` when the "Next" form nav button is clicked (`page_index` is the _destination_ page index, starting from zero)
- `form.wizardNavigationClicked` when the page nav button is clicked, but _before the page is set_ (`page_title` is the title of the selected page)
- `form.wizardPageSelected` _after_ the page nav is clicked and the page is selected (`page_index` and `page_title` are the index and title of the selected page)
- `form.saveDraft` when a draft submission is saved (no additional data)
- `form.restoreDraft` when a draft submission is restored (no additional data)
- `form.editGridAddRow` when the "Add another" button is clicked in an edit grid (`component` is the component key)

PR #740, opened 2020-11-18
SG-1235 - update template and css for form page content type

aekong:


web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -59,7 +59,7 @@
 (function() {
 
   var time = Date.now()
-  var confirmationURL = {{ paragraph.field_formio_confirmation_url.value | escape('js') }}
```
aekong, line 62: @nicolesque i think this was part of #736, the amplitude event stuff for forms. https://github.com/SFDigitalServices/sfgov/pull/736/files#diff-d5bc080583607c0222ee1f05139afed9692a1a185a19f4c2d0152e5405732df3R62.  the missing double quotes breaks forms that have a confirmation page url, like this one: https://pr-735-sfgov.pantheonsite.io/node/1316.  This has already been merged into `develop` so we can't deploy until either 1) this pr is merged to `develop` or 2) we create another pr to fix the missing double quotes.  just a heads up

jacine: Great catch @aekong! Notice this issue earlier on another env, where the forms w/conf urls don't load at all in Safari.

web/themes/custom/sfgovpl/templates/layout/page--node--form-page.html.twig
```diff
@@ -0,0 +1,96 @@
...
+  {% if page.alpha %}
+    {{ page.alpha }}
+  {% endif %}
+
+  <header role="banner">
+    <div class="sfgov-nav-container">
+      <div class="sfgov-nav-container__content">
+        {{ [page.header.sfgovpl_branding] }}
+        <div class="head-right--container">
+          {{ page['#title']|raw }}
```
zakiya, line 62: Will 'page.label' work? I like to avoid '|raw' when possible.

zakiya (approved):

PR #742, opened 2020-11-25
[SG-1262] Campaign spotlight using yellow theme should have slate blu…

jacine:
…e text.

This is actually for SG-1262! Whoops.


minnur (approved):
LGTM

PR #744, opened 2020-12-01
Sg 1089 homepage reading order

aekong:


config/views.view.services.yml
```diff
@@ -501,12 +501,10 @@ display:
         css_class: false
         title: false
       row:
-        type: fields
+        type: 'entity:node'
```
jacine, line 504: 👏

web/themes/custom/sfgovpl/templates/components/topic-card.twig
```diff
@@ -1,6 +1,11 @@
+{% if elem %}
+  {% set el = elem %}
+{% else %}
+  {% set el = 'h3' %}
```
jacine, line 4: FYI, for future ref (no need to change now)... Better way to write this is: 

`{% set el = elem|default('h3') %}`, or just `{% set el = el|default('h3') %}`

web/themes/custom/sfgovpl/templates/node/node--topic--card.html.twig
```diff
@@ -4,12 +4,14 @@
...
-      } %}
-    </div>
+    {% if is_front %}
+      {% set el = 'p' %}
+    {% endif %}
+    {% include "@theme/topic-card.twig" with {
+      "title": node.label,
+      "description": content.field_description.0['#text'],
+      "url": url,
+      "elem": el ? el : ''
```
jacine, line 14: Can skip the declaration above too, and just do: 

```suggestion
      "elem": is_front ? 'p' : '',
```

jacine (approved):
Nice cleanup! @aekong  Love to see fields _just_ being printed in entity templates! 👍 

Left some FYI comments, no need to make changes.

PR #747, opened 2020-12-01
send slack notification as attachment to auto collapse

aekong:
| before | after |
|---|---|
| image | image |
| image | image|


jacine (approved):
This is awesome! I _need_ to know how you are testing this locally. 🤣

aekong:
Thanks for looking! Nothing very elegant, just `lando drush scr` or just `curl`-ing 😆

PR #753, opened 2020-12-04
handle empty array value for confirmation url

aekong:
for empty field values, drupal appears to send an empty array.  this was causing issues with the confirmation page redirect logic.

some notes copied from a slack convo:

> i think it has to do with this line: https://github.com/SFDigitalServices/sfgov/blob/f0d02537e37baf4da696815688fd51566cc5e158/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig#L62
>
>if paragraph.field_formio_confirmation_url.value doesn’t have a value, the twig filters evaluate to [] (empty array).
>
>further down the file:
>https://github.com/SFDigitalServices/sfgov/blob/f0d02537e37baf4da696815688fd51566cc5e158/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig#L200-L202
>
> since confirmationURL is `[]` and not `null` or `undefined`, `window.location = []` just refreshes the browser

### DIFF
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -59,7 +59,11 @@
 (function() {
 
   var time = Date.now()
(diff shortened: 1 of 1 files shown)


shawnbot (approved):
Good lord, I'm sorry for opening up this can of worms. This LGTM!

PR #754, opened 2020-12-04
bump formio-sfds to 7.0.2

shawnbot:
[This release](https://github.com/SFDigitalServices/formio-sfds/releases/tag/v7.0.2) fixes some missing form translation keys in our radio/checkbox templates. I've already updated the config in prod because it's needed for a form that's going live today.


jacine (approved):
Thanks @shawnbot! 👍

PR #763, opened 2020-12-15
[SG-1278] Allow for Form.io library versions to be manipulated by query strings in Drupal

jacine:


arlina-espinoza (changes requested):
Looks good and works correctly. The only change I suggest is updating the config form to remove the reference to the file `sfgov_formio.libraries.yml`, as it will be removed by this PR:

image

arlina-espinoza (approved):
LGTM, this is a clever solution :100: Thanks @jacine !

shawnbot (approved):
Thank you, huzzah!

PR #767, opened 2020-12-17
SG-1272 - add new tokens for amplitude tracking of related topics and depts

aekong:


jacine (approved):
LGTM! Thanks @aekong!

PR #768, opened 2020-12-22
Fix and centralize code for getting $node object in theme hooks…

jacine:
for all kinds of routes including:

- node view (necessary for non-node hooks, like html, page, and suggestions)
- revisions
- previews
- public previews
- diffs (coming soon)

_Note:  This code was split out from #739, which is getting some design changes, because it is needed for #766._


aekong (approved):
tested in pr-768, looks good!

arlina-espinoza (approved):
Really good cleanup @jacine , thanks!!

PR #771, opened 2021-01-05
[SG-1280] Update filter format on field_alert_text.

jacine:


web/modules/custom/sfgov_alerts/sfgov_alerts.install
```diff
@@ -0,0 +1,34 @@
...
+  $nodes = Node::loadMultiple($nids);
+
+  /** @var \Drupal\node\Entity\Node $node */
+  foreach ($nodes as $node) {
+    // Update format, retaining existing content as value.
+    $node->field_alert_text->setValue([
+      'value' => $node->field_alert_text->value,
+      'format' => 'sf_restricted_html',
+    ]);
+
```
jacine, line 28: @minnur I need your help with this one!  I have no idea why, but it's not working.  If I take this code and run it in Devel PHP, it works just fine. 😭

zakiya (approved):
Looks ok to me.

PR #774, opened 2021-01-08
Track form load errors, guard against ad blockers

shawnbot:
**UPDATE**: I'm hijacking this PR to also fix the issue of forms failing when ad blockers prevent Amplitude's JS from loading. The technique is to have the `amp()` function to log to the console _unless_ `window.amplitude` is truthy, in which case it swaps out the console logger for `window.amplitude.getInstance().logEvent()`.

As discussed with @jacine, @zakiya, and @nicolesque [in Slack](https://sfdigitalservices.slack.com/archives/C010PJ7HJR3/p1610058130031400), we don't currently have any way of tracking when forms completely fail to load from form.io. This PR attempts to address that by catching the promise that rejects when the form fails to load and fires a (newly minted) `form.loadError` event to Amplitude.

We should be able to test this by adding a form page with a bad data source URL—either a non-existent or malformed URL, or an existing form.io form or resource that's not publicly visible.

### DIFF
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -79,10 +79,19 @@
       })
     }
 
-    var amplitude = window.amplitude.getInstance()
-    var amp = function(eventName, data) {
-      amplitude.logEvent(eventName, Object.assign(data, amplitudeData))
-    }
+    var amp = (function() {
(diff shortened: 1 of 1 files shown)


zakiya (approved):
I don't see any issues, but I'm not that familiar with this product.

aekong (commented):
nice.  is this still in draft? or ready to go?

aekong:
🤞

aekong (approved):

PR #781, opened 2021-01-13
Update Lando Configuration for Xdebug 3 and Composer

jacine:
A [new version](https://xdebug.org/docs/upgrade_guide) of Xdebug came out last month.  The next time you rebuild your Lando instance, you will get v3. v2 settings are not compatible with v3.  You would also get Composer 2, which will not work with our code base.  This PR:

1. Updates the Xdebug settings for v3.
2. Locks Lando's composer version to 1.x.
3. Removes Twig Xdebug module. Twig tweak actually provides the ability to use `{{ drupal_breakpoint() }}`, which is basically the same thing.

We can/should still use the tooling commands to toggle Xdebug, e.g. `lando xdebug-on`.

---

## Instructions

1. Rebuilding your Lando instance is required: `lando rebuild -y`
2. If you use VSCode, you'll need to update the `launch.json` config. Specifically, the default port is now 9003. Here's my configuration:

```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Listen for Xdebug",
      "type": "php",
      "request": "launch",
      "port": 9003,
      "log": true,
      "pathMappings": {
        "/app": "${workspaceRoot}",
      }
    }
  ]
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,7 +111,8 @@ Thumbs.db
 !web/core/**/*.gz
 
 # visual code config
-.vscode
+.vscode/*
+!.vscode/launch.json.example
 
 # secrets
 # These are created in the build process by circleci
--- a/.lando.yml
+++ b/.lando.yml
@@ -4,12 +4,12 @@ config:
   framework: drupal8
   site: sfgov
   id: 91d50373-c4cf-40e4-a646-cb73e16a140c
-  composer_version: 1-latest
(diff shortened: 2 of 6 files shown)


zakiya (approved):
Onward!

zakiya:
I added the sample .vscode file to the repo and I can also confirm this configuration works with PHPStorm.

zakiya (approved):

PR #782, opened 2021-01-13
SG-1303 - show transcript toggle only if transcript exists

aekong:


zakiya (approved):
Looks good!

aekong:
Thanks for looking!

PR #784, opened 2021-01-19
Add {lang} placeholder support in confirmation URL

shawnbot:
This adds support for a `{lang}` placeholder in form page confirmation URLs. I updated the field description, so now it looks like this when you're editing:

image

Here's an acceptance test:

```feature
Feature: form page confirmation URL

  Scenario: It works without a "{lang}" placeholder
   Given a form page
     And the confirmation URL is "/my-form/confirm"
    When I submit the form
    Then the URL should be "/my-form/confirm"

  Scenario: It works with a "{lang}" placeholder
   Given a form page
     And the confirmation URL is "/{lang}/my-form/confirm"
     And the page is translated in Spanish
    When I visit the form in Spanish
     And I submit the form
    Then the URL should be "/es/my-form/confirm"
```

cc: @nicolesque

### DIFF
--- a/config/field.field.paragraph.form_io.field_formio_confirmation_url.yml
+++ b/config/field.field.paragraph.form_io.field_formio_confirmation_url.yml
@@ -15,7 +15,9 @@ field_name: field_formio_confirmation_url
 entity_type: paragraph
 bundle: form_io
 label: 'Confirmation Page URL'
-description: 'Specify a confirmation page URL, if desired. Use the <a href="/node/add/form_confirmation_page" target="_blank">form_confirmation_page</a> content type to add a new confirmation page.'
+description: |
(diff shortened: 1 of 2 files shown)


aekong:
change looks fine and the additional acceptance test is icing on the cake 🍰 !

but where are we executing the acceptance test? i don't believe we have anything set up to run these tests during build

aekong:
ah cool, thanks for keeping it on the radar.  will add that to the list of things for us to connect on

aekong (dismissed):

aekong:
should be here: /admin/structure/paragraphs_type/form_io/fields/paragraph.form_io.field_formio_confirmation_url

aekong (approved):
🚢

PR #788, opened 2021-01-21
add media entity file replace and file delete modules

aekong:
https://www.drupal.org/project/media_entity_file_replace
https://www.drupal.org/project/file_delete


config/core.entity_form_display.media.file.default.yml
```diff
@@ -26,13 +26,13 @@ mode: default
 content:
   created:
     type: datetime_timestamp
-    weight: 10
+    weight: 11
```
nicolesque, line 29: @aekong why'd you bump these weights?

aekong: the file replace field wasn't enabled by default.  bumped it up to right under the `file` field, otherwise, would have been at the bottom and a bit out of context

nicolesque (approved):
LGTM! Merge and deploy to test!

PR #789, opened 2021-01-21
Alias "zh-TW" and "zh" to "zh-hant" in form confirmation URLs

shawnbot:
This adds a mapping of the `zh-TW` and `zh` language codes to `zh-hant` in form confirmation URLs with the `{lang}` placeholder to work around inconsistencies between Drupal, Google Translate, and Phrase.

During the first pass at testing I noticed that we don't support `/en/`-prefixed URLs, so in 1347b1e I opted to strip the `{lang}/` bit from the confirmation URL if the language is `en`.

@aekong LMK if you think this should live somewhere else. It might be good to have all of our language code mappings in one place, but I might need your help figuring out where that would be so that they're available to all of the scripts that need them. These are the other two I found:

https://github.com/SFDigitalServices/sfgov/blob/4fa402afe2c5e1badd2b71771d888e80708b2470/web/themes/custom/sfgovpl/src/js/translate.js#L168-L170

https://github.com/SFDigitalServices/sfgov/blob/1d246f67ff47c0745d01474a89c6aa80698daa6a/web/themes/custom/sfgovpl/includes/html.inc#L133-L136

There is one other [mapping](https://github.com/SFDigitalServices/formio-sfds/blob/b2acc4e1ea0297021975ab316a6de73ab332ba9c/src/patch.js#L361-L364) in the formio theme, but it's specific to Flatpickr, which is much pickier (sorry) about language codes. 🙃

### DIFF
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -223,7 +223,7 @@
             // see: <https://github.com/formio/formio.js/pull/3592> for more details
             var formLang = form.language || form.options.language
             var lang = formLanguageMap[formLang] || formLang
-            window.location = lang
+            window.location = lang && lang !== 'en'
               ? confirmationURL.replace('{lang}', lang)
               : confirmationURL.replace('{lang}/', '')
           }
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -66,6 +66,11 @@
       : paragraph.field_formio_confirmation_url.value|json_encode(constant('JSON_PRETTY_PRINT'))|raw
   }}
 
+  var formLanguageMap = {
+    zh:       'zh-hant',
+    'zh-TW':  'zh-hant'
+  }
+
   window.addEventListener('load', function() {
     var loadTime = Date.now() - time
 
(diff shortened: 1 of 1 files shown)


aekong:
sounds like drupal is the common thread here, so it sounds like it should live there.

maybe we should expose the single language map to `drupalSettings.sfgov_translations` so that it's available to all the front end js.

as for the `/en/{url}` we should probably have a redirect for this to `/{url}`

aekong (approved):

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -66,6 +66,11 @@
       : paragraph.field_formio_confirmation_url.value|json_encode(constant('JSON_PRETTY_PRINT'))|raw
   }}
 
+  var formLanguageMap = {
```
nicolesque, line 69: Does it seem sus that we have to do this? I'm fine with it for now but doesn't seem ideal. I feel like we should handle this more holistically rather than making mappings in any given template cc: @aekong

nicolesque:
I tested using Shawn's test form and it looks good to me. @aekong you can merge and deploy to test as soon as you're able.

nicolesque (approved):

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
aekong: yeah, shawn and i discussed.  since there seems to be some urgency around this, we can push this forward and come back to make the language mapping more accessible to other scripts as part of the `drupalSettings` object

PR #790, opened 2021-01-21
Log form submission metadata in amplitude

shawnbot:
This adds three keys to the form `submitDone` event handler that generates data for Amplitude:

1. `id` is the unique submission ID (`_id` because Mongo)
2. `form` is the unique form ID
3. `project` is the unique project ID

This should allow us to get more information about odd form submissions and figure out where invalid autocomplete values are coming from (browser, version, user journey, etc.).

🔒 **None of the submission data is included in this payload.**

### DIFF
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -111,7 +111,11 @@
         return { } // just log the event
       },
       submitDone: function (submission) {
-        return { } // just log the event
+        return {
+          id: submission._id,
+          form: submission.form,
+          project: submission.project
+        }
       },
       error: function (message) {
         return { error: message }


aekong (approved):

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -111,7 +111,11 @@
         return { } // just log the event
       },
       submitDone: function (submission) {
-        return { } // just log the event
+        return {
```
shawnbot, line 114: I'm _fairly_ confident that the `submission` object will always be defined here, but I guess we could guard against that with something like:

```suggestion
        return (submission instanceof Object) && {
```

What do you think, @aekong? We're already wrapping these calls in a try/catch, so if it throws on `submission._id` then we just won't get a `submitDone` event in Amplitude.

aekong: 👍

PR #812, opened 2021-02-10
Build theme assets in CI; add Babel for compiling JS

shawnbot:
With some great guidance from @zakiya, I've made a couple of changes to our sfgovpl library setup that should make it a bit easier to do some other things in the future. Here's the deal:

#### `sfgovpl/common` library
There is now an `sfgovpl/common` library definition that bundles up all of the `sfgovp/*` libraries that we previously listed under `libraries` in `sfgovpl.info.yml`. In 9a6579a @zakiya added a Drupal system performance config that enables asset minification + concatenation during development, which gave us a home in git for our cache max-age of 15 minutes.

#### `.browserslistrc`
I've deleted `config.js` in favor of inlining the SCSS input and output paths in `gulpfile.js` and moving the browser list to [`.browserslistrc`](https://github.com/browserslist/browserslist). This file is discoverable by lots of other tools. There weren't any references to the `images` key in the deleted config.

#### JS build step
In #814 I added a JS build step with Babel and the ["env" preset](https://babeljs.io/docs/en/babel-preset-env), so we can write modern JS syntax (`const`, `for..of`, [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), etc.) and support all of the browsers in our list without thinking about it.

#### Build assets on Circle
In #814 I also added an asset build step on Circle and updated the libraries config to reference only files in the `dist/` directory. Then I 🔥 deleted the `dist/` directory from git so we don't need to commit those files, because that's error-prone and very likely to produce merge conflicts.

#### Prettier
Per @zakiya's suggestion, I've also brought in [Prettier](https://prettier.io/), which can help us format _and_ lint JS, CSS, and even [Twig templates](https://github.com/trivago/prettier-plugin-twig-melody)! For now, you can run it manually with:

    ```sh
    # list files with formatting errors
    npx prettier --check src/js/whatever.js 

    # reformat files in place
    npx prettier --write src/js/whatever.js

    # checking templates currently outputs a lot of parse errors! 😬 
    npx prettier --check templates
    ```

In a future PR we can set up linting in CI and/or with [precommit hooks](https://prettier.io/docs/en/precommit.html) so that it's (basically, mostly) impossible for people to commit badly formatted code.


web/themes/custom/sfgovpl/gulpfile.js
```diff
@@ -8,39 +6,40 @@ const autoprefixer = require('autoprefixer');
 const browsersync = require('browser-sync').create();
 const path = require('path');
 
-const config = require('./config');
-const drupalLibraries = path.resolve(__dirname, '../../../libraries'); // drupal libraries which include sass source files
+// drupal libraries which include sass source files
+const drupalLibraries = path.resolve(__dirname, '../../../libraries');
```
zakiya, line 10: We should evaluate whether this line is necessary at some point. It might be a vestige from the pattern library days. The only thing in libraries right now is jquery.instagramfeed which is added through Drupal's library system and so shouldn't be needed in the theme build.

zakiya:
Thanks for doing this! The PR looks good!

However - CSS and JS aggregation is not enabled on live. I wouldn't have thought to check but you mentioned there being a ton of js files and I wondered. Do you want me to investigate and add that setting to this PR? @aekong @shawnbot

aekong:
Hey @zakiya thanks for looking. Yeah please have a look. I thought I had config split it so that we could ignore import/export. Was trying to leave the setting on prod but keep our local envs with aggregation off

zakiya:
With the existing config.split setting, the css and js aggregation are be disabled on configuration import (drush cim). I added that file back to the config folder so aggregation will be enabled on all environments unless explicitly changed in (*.)settings.php. 

To disable css and js on your local, use ```$config['system.performance']['css']['preprocess'] = FALSE;
$config['system.performance']['js']['preprocess'] = FALSE;``` in your settings.local.php. Note that the checkboxes will still appear as enabled at /admin/config/development/performance even when disabled through the $config array.

web/themes/custom/sfgovpl/gulpfile.js
shawnbot: You're right! I'm going to delete it.

config/system.performance.yml
```diff
@@ -0,0 +1,7 @@
+cache:
+  page:
+    max_age: 0
```
shawnbot, line 3: @zakiya is this right? Does this map to the max age in cache-control headers, or is this something else?

zakiya: Yes, this is the same as max-age in Cache-Control headers. @aekong Can you confirm that this should be 0? That's what it's set to on production.

aekong: it should be 15.  it was 0 in production because 15 wasn't in the config and got wiped out on the last release.  thanks 🙏

shawnbot, line 3:
```suggestion
    max_age: 15
```

shawnbot: Cool, I bumped it to 15 in 0114c71.

zakiya: The value is in seconds should be 900 if it's supposed to be 15 minutes.

shawnbot: Thank you! Good catch. 😬

config/system.performance.yml
```diff
@@ -0,0 +1,7 @@
+cache:
+  page:
+    max_age: 15
```
shawnbot, line 3:
```suggestion
    # seconds (15 * 60 = 900)
    max_age: 900
```

web/themes/custom/sfgovpl/gulpfile.js
```diff
@@ -1,46 +1,55 @@
...
-  browsersync.init({
+  return browsersync.init({
     proxy: 'https://sfgov.lndo.site/'
-  });
+  })
 }
 
 function watch() {
-  gulp.watch(config.css.source, css);
+  return gulp.parallel(
```
aekong, line 51: returning `gulp.parallel` here gave me an assertion error `Task never defined: [object Object]`.  Omitting the return and just executing the `gulp.watch` statements work, but breaks the pattern of returning for tasks.

Alternatively, we could break up the js and css watch tasks into two separate tasks that return `gulp.watch` and then throw them in the calls to `gulp.parallel` on lines 15 and 16.

what do you think @shawnbot ?

shawnbot: It looks from [the docs](https://gulpjs.com/docs/en/api/parallel/) like `watch()` shouldn't be a function, but should just be:

```js
const watch = gulp.parallel(
  // ...
)
```

I honestly don't know why some things are functions and some aren't, but maybe this is the pattern we should be following?

aekong: some are functions to allow for individual tasks and re-use in larger build tasks.  but also seems to be the pattern when using named functions for gulp.series/parallel? at least that seems to be what's indicated by those docs in the section about "[Avoid duplicating tasks](https://gulpjs.com/docs/en/api/parallel#avoid-duplicating-tasks)."

for watch specifically, yeah it doesn't need to be a separate function.

On a separate thing, it seems like browsersync isn't auto reloading the browser.  Would you mind if I made a pr to this branch to try and get it working?

aekong:
@zakiya this is looking good on my end.  our development workflow doesn't really change - we'll still have to `gulp` inside the `/web/themes/custom/sfgovpl` directory.

could you give this a quick review to be sure that it doesn't interrupt your work with vaccines?

thanks!

zakiya (dismissed):
Let's merge!

aekong:
i'm going to resolve that conflict and merge.  we've all touched it at some point and i think we're all in agreement that it looks good! thanks @shawnbot and @zakiya !

aekong (dismissed):

PR #814, opened 2021-02-10
Add Babel for JS compilation, build assets in CI

shawnbot:
This builds on #812 and adds Babel with the "env" preset to compile all of our theme's JS source files (`src/js/*.js`) into something that all of the browsers listed in our new `.browserslistrc` can parse. This means that we can use lots of modern JS (`cost`, `for..of` loops, object and array destructuring, and most modern _syntax_ features) right now, but we will need to do a bit more work to incorporate polyfills for modern browser APIs, including useful functions like `Object.assign()` and new Array prototype methods.

As a test for this setup, I've rewritten `components.js` to use DOM methods instead of jQuery. Let's test the step-by-step pages on old browsers with this new JS and see if it works!

I've also brought in [standard](https://standardjs.com/) as a Node dependency so that we can start linting our JS source files. No linting happens in CI now, but we can add that in a future PR. For now, you should be able to "fix" existing files with:

```sh
npx standard --fix src/js/whatever.js
```

~@zakiya you mentioned that we commit dist files (compiled CSS) into git right now, so we might have to do this with `dist/js/*.js` for now, but I would like to solve that problem soon, too. Could we ignore the `dist` directory during development, but then "un-ignore" it when pushing to Pantheon?~ 🤔 Figured out in #816, merged into this branch.


zakiya:
The [current pantheon docker image has node](https://github.com/pantheon-systems/docker-build-tools-ci) so I believe we can try building js and css on push.  If we do that, we should add node to the lando config as well. 

I'm all for using modern js over jquery but we should be especially aware of any changes to code using [Drupal.behaviors](https://drupalbook.org/drupal/68-working-javascriptjquery-drupal-8-what-are-behaviors)  or `drupalSettings`. Working without behaviors makes me nervous but that may be my Drupal Stockholm Syndrome talking and I see it's been done quite a bit already. 

I haven't used standardjs before but it looks like it can be used with [Prettier](https://github.com/prettier/prettier), which I am a big fan of. I've been using it locally to autoformat on save for this project and we've used it in precommit hooks for other big projects.

zakiya:
Guessing this is work in progress but there aren't any dist/js files in the repo so this shouldn't be merged until they're added or we add those build steps.

web/themes/custom/sfgovpl/gulpfile.js
```diff
@@ -1,34 +1,44 @@
...
-  const plugins = [
-    autoprefixer()
-  ];
   return gulp
     .src('./src/sass/**/*.scss')
     .pipe(sourcemaps.init())
     .pipe(sass({ 
-      outputStyle: 'expanded',
-      includePaths: [drupalLibraries]
+      outputStyle: 'expanded'
```
shawnbot, line 37: We can shrink the CSS in production with:

```suggestion
      outputStyle: process.env.NODE_ENV === 'production'
        ? 'compressed'
        : 'expanded'
```

Or we could add [cssnano](https://cssnano.co/) to our postcss setup.

web/themes/custom/sfgovpl/gulpfile.js
```diff
@@ -41,5 +51,5 @@ function serve() {
 }
 
 function watch() {
-  return gulp.watch(config.css.source, css);
+  return gulp.watch('./src', assets);
```
shawnbot, line 54: Currently this rebuilds both the JS and CSS whenever a change happens in `src`. We could be smarter about this with, say:

```suggestion
  return gulp.parallel(
    gulp.watch('./src/js', js),
    gulp.watch('./src/sass', css)
  );
```

web/themes/custom/sfgovpl/prettier.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  arrowParens: 'avoid',
+  semi: false,
+  singleQuote: true,
+  tabWidth: 2,
+  trailingComma: 'none'
```
shawnbot, line 6: These are very much my personal preferences, though they align pretty closely with [standard's rules](https://standardjs.com/rules.html). I think there's a separate convo to have about code style and linting—prettier only does the former; I think linting is equally, if not more important—but since any of these tools can rewrite the source files, we can just cross that bridge when we get to it.

web/themes/custom/sfgovpl/src/js/components.js
```diff
@@ -1,18 +1,23 @@
...
+  if (!trigger) {
+    console.error('Toggle element contains no trigger (with the "data-toggle-trigger" attribute)', toggle)
+    continue
   }
-})(jQuery);
+
+  trigger.addEventListener('click', e => {
+    const triggerLink = e.target
+    const showMoreText = toggle.getAttribute('data-show-text') || 'Show more'
+    const showLessText = toggle.getAttribute('data-hide-text') || 'Show less'
```
shawnbot, line 14: It would be great if we could figure out how to localize these strings. This change makes it possible to translate them in the template, e.g. assuming `t()` is a translation function:

```twig
<button
  data-show-text="{{ t('Show more') }}"
  data-hide-text="{{ t('Show less') }}">
```

web/themes/custom/sfgovpl/prettier.config.js
zakiya: I didn't mean to suggest I was against linting. I think Prettier and linting are complimentary, not mutually exclusive. 

I don't have strong feelings about format rules except for setting `twigOutputEndblockName: true` when using  https://github.com/trivago/prettier-plugin-twig-melody

shawnbot: Understood. It's great that there's a Twig plugin for Prettier! Let's set that up in another PR.

shawnbot: Maybe [prettier-standard](https://github.com/sheerun/prettier-standard) is the thing??

web/themes/custom/sfgovpl/src/js/components.js
zakiya: Can you clarify what you mean by localize?

This is how to use the t function in Drupal twig:
```<button
  data-show-text="{{ 'Show more'|t }}"
  data-hide-text="{{ 'Show less'|t }}">
```

shawnbot: I went back and looked at the templates and saw that, thank you! I edited the above comment _just_ before you replied. 😊

shawnbot: Done in 615d6fb7!

web/themes/custom/sfgovpl/gulpfile.js
shawnbot: Done in a953aa8ab48afccd584b0d59b68bd95d96ea788e.

web/themes/custom/sfgovpl/src/js/components.js
```diff
@@ -1,18 +1,24 @@
...
+for (const toggle of toggles) {
+  const trigger = toggle.querySelector('[data-toggle-trigger]')
+
+  if (!trigger) {
+    console.error('Toggle element contains no trigger (with the "data-toggle-trigger" attribute)', toggle)
+    continue
   }
-})(jQuery);
+
+  const showMoreText = toggle.getAttribute('data-show-text') || toggle.textContent || 'Show more'
```
aekong, line 11: i think you can also use `Drupal.t(string)` for localization on the js side

```suggestion
  const showMoreText = toggle.getAttribute('data-show-text') || toggle.textContent || Drupal.t('Show more')
```

@zakiya confirm?

web/themes/custom/sfgovpl/package.json
```diff
@@ -18,6 +18,7 @@
     "gulp-sass": "^4.1.0",
     "gulp-sourcemaps": "^2.6.5",
     "node-sass": "^4.14.1",
-    "prettier": "^2.2.1"
+    "prettier": "^2.2.1",
+    "prettier-plugin-twig-melody": "^0.4.6"
```
shawnbot, line 22: @zakiya this totally works! 🎉 

```twig
sfgovpl % npx prettier -c templates/paragraphs/paragraph--process-step.html.twig

Checking formatting...
templates/paragraphs/paragraph--process-step.html.twig[error] templates/paragraphs/paragraph--process-step.html.twig: Error: ERROR: Invalid Token
[error]   51 | {%
[error]   52 |   set showMoreText = 'Show more'|t
[error] > 53 |   set showLessText = 'Show less'|t
[error]      |   ^^^
[error]   54 | %}
[error]   55 | {% block paragraph %}
[error]   56 |   <div{{ attributes.addClass(classes) }} data-toggle-container>
[error] 
[error] Expected tag end "%}" but found symbol instead.
```

web/themes/custom/sfgovpl/src/js/components.js
shawnbot: Thanks, but I opted to do this in the template. We only need to translate each of those strings once for the entire page at this point, but we could just move the `t` calls inline if our needs change in the future.

PR #815, opened 2021-02-11
[SG-1316] redirect on alias

zakiya:


minnur (approved):
I think this looks good, I haven't done actual testing with the running Drupal.

aekong (approved):

PR #829, opened 2021-02-22
Sg 1315 translate people

aekong:


zakiya (approved):
Awesome. I tested and it works.

PR #833, opened 2021-02-23
[SG-1296] Fix height and margin for resource card

arshad:
Fixes [SG-1296]

[SG-1296]: https://sfgovdt.jira.com/browse/SG-1296


zakiya (approved):
👍

PR #837, opened 2021-02-23
prevent unnecessary clone content on subsequent pr builds

aekong:


zakiya (approved):
Nice!

@arshad and @minnur FYI - PR environments won't pull  a new database with each code push once this is merged.

PR #850, opened 2021-03-05
[VAX-113] - Count all sites, not just ones that match filter.

zakiya:
https://pr-850-sfgov.pantheonsite.io/vaccine-sites


aekong (approved):
🚀

PR #851, opened 2021-03-05
Vax 94 amplitude

aekong:
This is a first pass to move our amplitude stuff to our own module.  We had to patch the existing drupal module so many times it doesn't really seem worthwhile to continue adding bloat to it in order to do the things we want.

This pr will satisfy https://sfgovdt.jira.com/browse/VAX-94 (tracking applied filters to /vaccine-sites).  After that, the plan is to migrate the events and other patches to this `sfgov_amplitude` module.

Also would like to consolidate the formio amplitude tracking here as well.


zakiya (approved):
Looks good too me.

PR #859, opened 2021-03-10
[SG-1331] Add no translate class to headings. (Re-up PR as patch)

arshad:
Go CI


zakiya (approved):
Ideally we would use an attributes array to assign the notranslate class wherever it's used but inside the markup is also fine.

PR #861, opened 2021-03-10
[testing] Test vaccine site ordering

shawnbot:
A la #845


nicolesque:
@shawnbot we can probably get rid of this, right? we have the config to change the URL on all envs now!

PR #880, opened 2021-03-19
Relocate saml dependencies

aekong:
This pr aims to do the following:
* remove saml configs from the repo
* copy the environment specific dependencies from the `files/private` directory on pantheon (not source controlled)


josh-chou (dismissed):
👍

PR #882, opened 2021-03-24
[VAX-165] location filter

zakiya:
- Requires files to be pulled from live!
- Do not merge until okay'd by Nicole.

### DIFF
--- a/web/modules/custom/sfgov_vaccine/js/filter_sites.js
+++ b/web/modules/custom/sfgov_vaccine/js/filter_sites.js
@@ -139,11 +139,14 @@
                 $(this).removeClass(class_match_radius);
               }
               $(this)
-                .find(".distance")
+                .find(".vaccine-site__distance")
                 .text(Math.round(distance * 10) / 10 + "mi");
               $(this)[0].setAttribute("data-distance", distance);
+              $(this)
+                .find(".vaccine-site__header")
+                .addClass("distance-visible");
             } else {
-              $(this).find(".distance").text("");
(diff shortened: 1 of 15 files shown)


web/modules/custom/sfgov_vaccine/js/filter_sites.js
```diff
@@ -159,6 +211,28 @@
         $(".vaccine-filter__sites").hide();
       }
 
+      // @see https://en.wikipedia.org/wiki/Haversine_formula
```
aekong, line 214: 🤯

web/themes/custom/sfgovpl/src/js/filters.js
```diff
@@ -1,81 +1,95 @@
-(function($, window, Drupal) {
-  'use strict';
+;(function ($, window, Drupal) {
```
aekong, line 1: is the leading semicolon a better practice for us (i guess for aggregating and minification purposes)?

aekong (dismissed):

web/themes/custom/sfgovpl/src/js/filters.js
zakiya: This came from prettier. I think because of the no semi colon setting. 

https://github.com/SFDigitalServices/sfgov/blob/develop/web/themes/custom/sfgovpl/prettier.config.js
https://prettier.io/docs/en/options.html#semicolons

aekong (approved):

PR #886, opened 2021-04-02
Design system

aekong:
This pr aims to integrate the new sfds design system into the drupal theme for sf.gov by:
* pulling in the design system npm package as a dependency in the `sfgovpl` theme
* including the dist css from the design system as a drupal library
* modifying individual templates to utilize the classes from the design system (primarily buttons)

#### Results of a first pass (some tweaking needed):
* https://pr-886-sfgov.pantheonsite.io/departments/city-administrator/office-civic-engagement-and-immigrant-affairs
  * spotlight buttons

    Screen Shot 2021-04-05 at 9 20 16 PM

* https://pr-886-sfgov.pantheonsite.io/
  * campaign paragraph "Learn more" button

    Screen Shot 2021-04-05 at 9 21 14 PM

#### Some observations:
* some templates are easier to deal with than others.  for example: [web/themes/custom/sfgovpl/templates/field/field--paragraph--field-link--campaign.html.twig](https://github.com/SFDigitalServices/sfgov/pull/886/files#diff-10580e1612294cf8c54f232cdf24cc11f48ce66a806c6cec93b4e72d196929ffR7) allows us to target a specific field and provide custom markup and apply the classes from the design system (we can probably further separate this by creating a link/button component and including it here).
* there are areas of the site where the links that we are styling as buttons are coming from a wysiwyg textarea:
  * Here we might want to actually put in the source sass files so that we can import the button style that way.  But curious what that looks like since we're using tailwind in the design systemScreen Shot 2021-04-02 at 10 57 28 AM
* ~~i'm also not sure how to get into the `children` twig var: https://github.com/SFDigitalServices/sfgov/blob/design-system/web/themes/custom/sfgovpl/templates/form/container--more-link--news--block-1.html.twig so that I can get to the actual markup and apply a `.btn` class to it.  it's one of those drupal thingies that the markup is buried in a render array that i have to dig for~~
  * @zakiya added some helpful info about how to apply a class/add an attribute to markup generated internally by drupal: 
      
      ``` 
      function sfgov_utilities_preprocess_views_view(&$variables) {
        if($variables['id'] == 'news' && $variables['display_id'] == "block_1" ) {
          $url = $variables['more']['#url'];
          $url->setOption(
            'attributes', ['class' => ['btn']]
          );
          $variables['more']['#url'] = $url;
        }
      }
      ```


zakiya:
@aekong Example of adding class `btn` to the views more link for news block_1. Function could go in the theme with a different prefix obviously.

```
function sfgov_utilities_preprocess_views_view(&$variables) {

  if($variables['id'] == 'news' && $variables['display_id'] == "block_1" ) {

    $url = $variables['more']['#url'];
    $url->setOption(
      'attributes', ['class' => ['btn']]
    );
    $variables['more']['#url'] = $url;
  }
}
```

aekong:
whew, thanks @zakiya!

web/themes/custom/sfgovpl/sfgovpl.libraries.yml
```diff
@@ -27,6 +28,15 @@ sfgov-drupal-specific-styles:
...
       'dist/css/drupal.css': {}
 
+# design system styles
+sfgov-design-system-styles:
+  css:
+    base:
+      'node_modules/sfgov-design-system/dist/css/base.css': {}
+      'node_modules/sfgov-design-system/dist/css/components.css': {}
+      'node_modules/sfgov-design-system/dist/css/sfds.css': {}
+      'node_modules/sfgov-design-system/dist/css/utilities.css': {}
```
shawnbot, line 38: For now, I think we only want the utilities and component (including button) styles:

```suggestion
      'node_modules/sfgov-design-system/dist/css/components.css': {}
      'node_modules/sfgov-design-system/dist/css/utilities.css': {}
```

This isn't documented anywhere, but:
* `sfds.css` is the whole shebang (base + utilities + components) and intended more for prototyping and static situations
* `base.css` is intended to be just the base styles that reset browser styles, establish base typography, and prime documents for utilities with stuff like `* { box-sizing: border-box; }`.
* `utilities.css` includes, as you've probably guessed, _only_ the Tailwind utilities.
* `components.css` includes our component styles, including buttons.

This is all open to discussion, of course! We may not need all of them, but I've found that separating them out and having a singular bundle can help when you're debugging and building examples in CodePen, where you might want to only use certain parts of the system.

web/themes/custom/sfgovpl/templates/field/field--paragraph--field-link--button.html.twig
```diff
@@ -0,0 +1,7 @@
+{% for item in items %}
+  {% include "@sfgov-design-system/button/button.twig" with {
+      "url": item.content['#url'],
+      "title": item.content['#title'],
+      "modifiers": ['btn-inverse']
```
shawnbot, line 5:
```suggestion
      "href": item.content['#url'],
      "content": item.content['#title'],
      "classes": ['btn-inverse']
```

web/themes/custom/sfgovpl/templates/field/field--paragraph--field-link--campaign.html.twig
```diff
@@ -0,0 +1,12 @@
+{#
+/**
+ * @file
+ * Removes field wrapper markup for fields on 'campaign' paragraph.
+#}
+{% for item in items %}
+  {% include "@sfgov-design-system/button/button.twig" with {
+      "url": item.content['#url'],
+      "title": item.content['#title'],
+      "modifiers": ['btn-inverse']
```
shawnbot, line 10:
```suggestion
      "href": item.content['#url'],
      "content": item.content['#title'],
      "classes": ['btn-inverse']
```

web/themes/custom/sfgovpl/src/sass/components/_spotlight.scss
```diff
@@ -80,18 +80,6 @@
...
-    padding: 13px 20px 11px;
-    background: $c-white;
-    color: $c-bright-blue;
-    border-radius: 8px;
-
-    &:hover {
-      color: $c-dark-blue;
-    }
-  }
-
```
shawnbot, line 94: YAS! Love seeing CSS deleted.

shawnbot (changes requested):
This is off to a great start. I have some suggestions about naming, which I'd be happy to discuss in Slack, during our DS meeting later today, or here!

web/themes/custom/sfgovpl/templates/components/sfgov-design-system/button/button.twig
```diff
@@ -0,0 +1 @@
+<a href="{{ url }}" class="btn {{ modifiers|join(' ') }}">{{ title }}</a>
```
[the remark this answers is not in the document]
aekong: ❤️

web/themes/custom/sfgovpl/templates/field/field--paragraph--field-link--button.html.twig
aekong: ❤️

web/themes/custom/sfgovpl/templates/field/field--paragraph--field-link--campaign.html.twig
aekong: ❤️

web/themes/custom/sfgovpl/templates/components/sfgov-design-system/button/button.twig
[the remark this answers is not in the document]
shawnbot: @aekong How do you feel about renaming this template to something like `button/link.twig` just to make it very clear (both when editing and when it's referenced in other templates) that it's rendering an `<a>`?

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--campaign-spotlight.html.twig
```diff
@@ -20,6 +21,15 @@
   }) %}
 {% endif %}
 
+{# Spotlight button link #}
+{% set has_spotlight_button = false %}
+{% if content.field_spotlight_button|render %}
+  {% set has_spotlight_button = true %}
+  {% set button_link = paragraph.field_spotlight_button.entity.get('field_link').get(0) %}
```
aekong, line 28: @zakiya This seems like a verbose and brittle way to get the actual field values of the `Button link` paragraph type, but it works so i went with it anyway 😬.  Suggestions welcome!

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--campaign-spotlight.html.twig
```diff
@@ -45,9 +55,13 @@
...
         {% endif %}
 
-        {% if content.field_spotlight_button|render %}
+        {% if has_spotlight_button %}
           <div class="sfgov-campaign-spotlight-button">
-            {{ content.field_spotlight_button }}
+            {% include "@sfgov-design-system/button/button.twig" with {
+                "href": button_url,
+                "content": button_content,
+                "classes": campaign_theme_label != 'blue' ? ['btn-inverse'] : []
```
aekong, line 63: @shawnbot the `blue` theme here doesn't get the white color text treatment because we inject critical css  in the `head` with this defintion:
```
a:link {
  color: #4f66ee;
}
```

wondering if we need to specify the tags for the styles in the design system: 
```
button, 
a.btn, 
.btn { 
  ...
}
```
not too excited about doing that, but curious what your thoughts are

example: https://pr-886-sfgov.pantheonsite.io/covid-19-vaccine-san-francisco

web/themes/custom/sfgovpl/src/sass/components/_spotlight.scss
aekong: 🙌

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--campaign-spotlight.html.twig
shawnbot: I assumed that nearly all of the button styles were `!important`, which _should_ override the critical CSS. But maybe not??

zakiya: I think we should avoid using !important if we can.  I developed [this mixin](https://github.com/SFDigitalServices/sfgov/blob/71ba64518f26ac10263192c187fed10a63bdda84/web/themes/custom/sfgovpl/src/sass/_mixins.scss#L206-L262) to address link colors in Drupal so that colors can be set exclusively by element specificity.  :link and certain helper classes are used in core and contrib and this accounts for everything (at least that was the intention). ` critical.scss` is pulling from that mixin among other thing so this will probably keep coming up unless either :link is removed from the mixin or the design system uses :link.

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--campaign-spotlight.html.twig
zakiya: You're probably going to run into the issue of referencing paragraphs within paragraphs a lot. I often reference [this file from @jacine ](https://github.com/chapter-three/jcc-srl/blob/master/web/themes/custom/atrium/templates/paragraph/paragraph--accordion-container.html.twig) as a great way to organize twig for this scenario. It might seem like overkill when there's only 1 sub paragraph but there's something to be said for recognizable code patterns. 

Since child paragraph types often vary on SF.gov, theoretically it might be useful to use `paragraph_parent` and `paragraph_child` as variables. I'd have to work on it to say for sure.

Also in line 28 `paragraph` is the same php class as `entity` so ` {% set button_link = paragraph.field_spotlight_button.entity.field_link.0 %}` will work.

I'd re-write the next few lines like:

```
  {% set button_link = paragraph.field_spotlight_button.entity.field_link.0 %}
  {% set button_content = button_link.title %}
  {% set button_url = button_link.url %}
```


In Drupal twig, "content" is usually a variable that includes a render array. If possible I would omit the word from the design system and use "text" "link" etc instead.

web/themes/custom/sfgovpl/templates/components/sfgov-design-system/button/button.twig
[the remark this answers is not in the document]
zakiya: Can you use "text" or "label" (or anything else) instead of "content"? "content" is a really loaded word on the Drupal side and I think it can cause confusion and/or unnecessary cognitive overhead.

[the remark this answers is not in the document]
aekong: ah yeah, thanks for the reminder.  `content` is an object used by twig templates (eg `content.field_spotlight_button` to access field names for rendering.  `text` or `label` implies that the only thing inside button markup would be plain text, but that may not always be the case (unless we standardize on that)?

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--campaign-spotlight.html.twig
shawnbot: I was wrong about the `.btn` styles getting `!important`. From [Tailwind](https://tailwindcss.com/docs/functions-and-directives):

> Any rules inlined with `@apply` will have !important removed by default to avoid specificity issues

You _could_ fix this specific issue with the `text-white` utility. However, that's a band-aid and probably doesn't scale. What about qualifying the `a:link` selector to exclude buttons, e.g. `a:link:not(.btn)`? `:not()` with simple selectors is [widely supported](https://caniuse.com/mdn-css_selectors_not).

web/themes/custom/sfgovpl/package.json
```diff
@@ -19,6 +19,7 @@
     "gulp-sourcemaps": "^2.6.5",
     "node-sass": "^4.14.1",
     "prettier": "^2.2.1",
-    "prettier-plugin-twig-melody": "^0.4.6"
+    "prettier-plugin-twig-melody": "^0.4.6",
+    "sfgov-design-system": "0.0.0-12f9409"
```
shawnbot, line 23: Let's test this with https://github.com/SFDigitalServices/design-system/pull/8:

```suggestion
    "sfgov-design-system": "0.0.0-708636a"
```

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--campaign-spotlight.html.twig
aekong: thanks @shawnbot , took your suggestion: https://github.com/SFDigitalServices/sfgov/pull/886/files#diff-543941fc70cb2275bef9c4f4ea41cd65f1bd7d02cb1b98e830d642d4c9a56974R53

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--campaign-spotlight.html.twig
aekong: thanks @zakiya took your suggestion: https://github.com/SFDigitalServices/sfgov/pull/886/files#diff-b83b66c4e77788772162ec9ea68685ac3fcb2e12e0a9fec96fe09722590a5d3aR28-R30

web/themes/custom/sfgovpl/templates/components/sfgov-design-system/button/button.twig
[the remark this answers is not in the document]
aekong: @zakiya i changed the property name to `text`

@shawnbot i renamed the template file to `button/link.twig`

web/themes/custom/sfgovpl/src/sass/paragraphs/_paragraph-campaign.scss
```diff
@@ -64,12 +64,3 @@
     margin-bottom: 40px;
   }
 }
-
-.sfgov-campaign-link {
-  margin-bottom: 20px;
```
shawnbot, line 69: @aekong I was looking at the `a {` deleted below and noticed that `sfgov-campaign-link` is only referenced once, and all it's doing is setting the bottom margin. So I replaced it with `mb-20` and deleted it! 🔥

aekong: github needs a :hell-yeah: reacji! 🔥 🔥 🔥

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--spotlight.html.twig
```diff
@@ -27,7 +27,7 @@
       {% endif %}
 
       {% if content.field_description|render %}
-        <div class="sfgov-spotlight-description">
+        <div class="mb-20 lg:mb-40">
```
shawnbot, line 30: This was another easy one to 🔥 . The CSS, for reference:

```scss
.sfgov-spotlight-description {
  margin-bottom: 20px;

  @include media($medium-screen) {
    margin-bottom: 40px;
    padding: 0;
  }
}
```

Our [`lg` breakpoint](https://github.com/SFDigitalServices/design-system/blob/535e422fd54fcdf3b811874f02c2da426b2c36b9/src/tokens/breakpoints.js#L2) is the equivalent of [`$medium-screen`](https://github.com/SFDigitalServices/sfgov/blob/ae0c49602678b97b132a29003827df1732f8060a/web/themes/custom/sfgovpl/src/sass/_variables.scss#L62) (768px).

web/themes/custom/sfgovpl/package.json
```diff
@@ -19,6 +19,7 @@
     "gulp-sourcemaps": "^2.6.5",
     "node-sass": "^4.14.1",
     "prettier": "^2.2.1",
-    "prettier-plugin-twig-melody": "^0.4.6"
+    "prettier-plugin-twig-melody": "^0.4.6",
+    "sfgov-design-system": "1.0.0"
```
shawnbot, line 23: [1.0.0](https://github.com/SFDigitalServices/design-system/releases/tag/v1.0.0) 🚀

web/modules/custom/sfgov_design_system/sfgov_design_system.module
```diff
@@ -0,0 +1,15 @@
...
+ */
+function sfgov_design_system_theme($existing, $type, $theme, $path) {
+  return [
+    'sfgov_design_system_components' => [
+      'variables' => [
+        'some_variable' => NULL,
+        'sfds_components' => NULL,
+      ],
+    ],
+  ];
```
shawnbot, line 14: What's the story here? What might we put in this array?

shawnbot: Oh I see, you're working on a components page! Disregard 😁

aekong: yeah, it's to set up the theme for the components page and the vars that get sent to the twig template: https://github.com/SFDigitalServices/sfgov/pull/886/files#diff-8a7732c7a680e436221b7109734ec32d9fc7227c23c76c17685c25246d96349bR24

what i have so far (which is not much) is here: https://pr-886-sfgov.pantheonsite.io/design-system/components

@zakiya and i had chatted briefly about the challenges trying to get the design system components within the context of standalone sf.gov components.  We may potentially miss things in the render pipeline like other classes or markup that get passed along (classes, additional markup).  

Also, the way the drupal template hierarchy is set up, we may need to re-work some things to better modularize the fields being rendered (for example the campaign spotlight: https://github.com/SFDigitalServices/sfgov/blob/develop/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--campaign-spotlight.html.twig) so that we can use those templates in this components page to see the design system components in better context.

shawnbot: Interesting, okay!

FWIW, I like the way you have the templates set up right now so that they accept "primitive" values and don't try to unpack Drupal objects. One thing I'm wondering is if we can take a page from the React book and call the content of a component `children` instead of `text` or `content` (per @zakiya's note above).

I'm not sure what the Right Way™ is to do this in Twig, but it would be cool if you could `apply` macros with the content passed as into the macro as the first argument, the same way that React works:

```jsx
<Component>
  foo <b>bar</b>
</Component>
```

The component gets `{children: <>foo <b>bar</b></>}` as props and can choose where to render them. I'm picturing something like:

```jinja
{% render link_button(href='...') %}
  This is the button text
{% endrender %}
```

But maybe that's over-thinking things? 🤔

shawnbot (approved):
I ran a Lighthouse audit locally for the [multidev](https://pr-886-sfgov.pantheonsite.io/) (as of ac22fc1, which is 12 days behind `develop` right now) and compared it with sf.gov. Here's what I noticed:

What | Production | This PR | Change |
:--- | :--- | :--- | :---
CSS transfer size (compressed) | 52.6K | 61.9K | +9.3K |
Unused CSS "potential savings" | 48.5K | 57.6K | +9.1K
Estimated savings | 0.16s | 0.2s | +0.04s

This is just one sample, but it's much better than I expected. The lion's share of the new CSS comes from [utilities.css](https://unpkg.com/browse/sfgov-design-system@1.0.0/dist/css/utilities.css), which weighs in at 54.3K. Utilities are very highly "compressible" because they contain lots of repeated strings, which is why adding it to our Drupal libraries config doesn't actually increase the Drupal CSS bundle by its size.

I would say that this is ready to go. We can optimize the CSS in another pass, for instance by generating a custom build of `utilities.css` with [PurgeCSS](https://purgecss.com/), which could delete any utilities (class selectors) that aren't mentioned in our Twig templates (and probably our JavaScript).

We can also build out the twig template examples in another pass. 😁

web/modules/custom/sfgov_design_system/sfgov_design_system.module
aekong: @shawnbot and i chatted briefly about this IRL - we'll address this possibility at a later date

shawnbot (approved):
SO EXCITED!

PR #887, opened 2021-04-02
SG-914 - set english as default language for admin content list

c3gabe:


zakiya (approved):
@c3gabe Just curious - was this not possible with config?

PR #893, opened 2021-04-12
[SG-1352] Accommodate end dates on events.

zakiya:


zakiya:
Closing since this PR is pretty stale and we need the environments.

PR #904, opened 2021-05-04
[SG-1387] QLess integration

zakiya:


config/sfgov_qless.settings.yml
```diff
@@ -0,0 +1,7 @@
+api_url: 'https://qless-microservice.herokuapp.com/api/v1/queues'
+error_message: 'We are having trouble reaching the service right now. Please try again later.'
+title: 'Wait Times'
+caption: 'The Permit Center uses Qless to manage customer lines and wait times.'
```
shawnbot, line 4: I don't know if we want to (or even can??) call them out by name here, but if so I think we should capitalize the "L" for brand consistency:

```suggestion
caption: 'The Permit Center uses QLess to manage customer lines and wait times.'
```

web/modules/custom/sfgov_qless/src/Form/SettingsForm.php
```diff
@@ -0,0 +1,130 @@
...
+   */
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+
+    $values = $form_state->getValues();
+    $url = $values['api_url'];
+    $format = UrlHelper::isValid($values['api_url']);
+
+    if ($format === TRUE) {
+      $request = $this->httpClient->get($url, ['http_errors' => FALSE]);
+      $code = $request->getStatusCode();
```
shawnbot, line 107: @zakiya I want to flag this because I'm currently returning `status: "error"` with a `200` status even if there are problems. We can handle non-200 statuses via XHR on the front end, so maybe errors should return a `500`?

zakiya: This is on the settings form to prevent people from entering a totally invalid url. Since that won't ever change in production, instead add an extra check on front end so we don't try to display the table if there's no data.

minnur (approved):
LGTM

PR #909, opened 2021-05-11
[SG-1378] Data Story

arshad:
Fixes SG-1378


web/themes/custom/sfgovpl/src/sass/node/_node-data-story.scss
```diff
@@ -0,0 +1,365 @@
...
+    top: 39px;
+  }
+
+  .toolbar-horizontal.toolbar-tray-open & {
+    top: 79px;
+  }
+
+  .sfgov-toc-tree {
+    flex: 1;
+    overflow: scroll;
```
shawnbot, line 355: Hi @arshad! This is causing some issues in all of the browsers on my machine:

Chrome, Edge, Safari | Firefox
:--- | :---
 | 

I believe the fix is:

```suggestion
    overflow-y: auto;
```

Generally I think we want to avoid `overflow: scroll` and use `auto` whenever possible. And unless there's a good reason for the content of this element to overflow (say, a very long word that can't break), I don't think we should let it scroll horizontally.

shawnbot (changes requested):
Please see my suggestion about `overflow: scroll`. 🤘

zakiya (approved):
The promoted and sticky fields should be hidden on the node edit/add page. I created SG-1426 for it. Other than that, the PR looks good.

PR #922, opened 2021-06-07
SG-1422 - hide campaign instagram field from form and template display

aekong:
Hides the instagram feed option from "Additional content" and "Social media embed" fields on the Campaign content type


zakiya (approved):
lgtm.

PR #925, opened 2021-06-22
remove xtm module

aekong:


zakiya (approved):
There's a composer conflict but other than that it looks good to go.

PR #927, opened 2021-06-23
configure writer role permissions

aekong:


config/user.role.writer.yml
```diff
@@ -7,32 +7,233 @@ label: Writer
...
+  - 'update media'
+  - 'use admin toolbar search'
+  - 'use editorial transition archive'
   - 'use editorial transition archived_draft'
+  - 'use editorial transition archived_published'
   - 'use editorial transition create_new_draft'
+  - 'use editorial transition publish'
   - 'use editorial transition submit_for_review'
+  - 'use text format basic_html'
+  - 'use text format full_html'
```
zakiya, line 198: I think things would go bonkers if we didn't included this but I want to flag that full_html is a gaping security hole and should be another high priority fix.

zakiya (dismissed):

zakiya (approved):

PR #928, opened 2021-06-23
update ctools to 3.7.0 + other module updates

aekong:
* Updates chaos tools to 3.7.0 to address security vulnerability: https://www.drupal.org/sa-contrib-2021-015
* also updates other modules


zakiya (approved):
⭐

PR #933, opened 2021-06-28
SG-1324 - make campaign top fact images more uniform

aekong:


zakiya:
I think this looks good. I tested on Chrome, Firefox, Safari - all screen sizes with two rows of facts.

web/themes/custom/sfgovpl/src/sass/paragraphs/_paragraph-facts.scss
```diff
@@ -52,51 +52,32 @@
...
+        strong {
+          font-weight: $fw-medium;
         }
       }
     }
   }
-}
+  @include media($medium-screen) {
+    display: grid;
+    grid-template-columns: 32% 32% 32%;
```
zakiya, line 79: try `grid-template-columns: 1fr 1fr 1fr;` instead

It should work with IE.

aekong: thx! updated

aekong: @zakiya the `1fr` unit gives me this on certain window sizes (i think it is accommodating for word wrapping):
Screen Shot 2021-06-29 at 1 24 58 PM

`32%` seems more what we want? seems to be a bit more consistent:
Screen Shot 2021-06-29 at 1 25 10 PM

zakiya: What's the url? If 32% works, I'm not against it but I've never had to specify a percent when using grid before.

aekong: sorry: https://pr-933-sfgov.pantheonsite.io/covid19youthservices

zakiya (approved):

PR #942, opened 2021-07-07
Use postcss as the main CSS processor

shawnbot:
This PR promotes [postcss](https://postcss.org/) to the job of top-level processor for our stylesheets. This is a step toward using [Tailwind](https://tailwindcss.com/) and inheriting the configuration from our design system so that we can more nimbly import different parts of the system and do things like:

- Include a specific subset of our utility classes in our `critical.css` bundle
- [Purge](https://tailwindcss.com/docs/optimizing-for-production#removing-unused-css) styles that aren't used in any of our templates
- Subset custom fonts more strategically (e.g. by deferring the bigger font payloads)

### What about Sass?
It's still in there! There's a handy [postcss plugin](https://www.npmjs.com/package/postcss-node-sass) that processes the (S)CSS source through node-sass, which allows us to do interesting things like process the source files before and after Sass with other postcss plugins (one of which we already use: autoprefixer!).


web/themes/custom/sfgovpl/gulpfile.js
```diff
@@ -1,16 +1,23 @@
...
-const sass = require('gulp-sass')
-const sourcemaps = require('gulp-sourcemaps')
-const postcss = require('gulp-postcss')
 const autoprefixer = require('autoprefixer')
+const babel = require('gulp-babel')
 const browsersync = require('browser-sync').create()
+const gulp = require('gulp')
 const path = require('path')
+const postcss = require('gulp-postcss')
+const sourcemaps = require('gulp-sourcemaps')
```
shawnbot, line 7: I alphabetized these so the diff is harder to read, but I've removed the `gulp-sass` import here. I also need to uninstall the npm dependency.

aekong (approved):
🍰

PR #945, opened 2021-07-07
[SG-1143] Permissions round 2

zakiya:


web/modules/custom/sfgov_moderation/src/StateTransitionValidation.php
```diff
@@ -0,0 +1,256 @@
...
+    if ($entity->getEntityTypeId() != 'node' ||
+      $entity->isNew() ||
+      $user->hasPermission('administer nodes') ||
+      $user->hasPermission('bypass node access') ||
+      in_array($this->publisher(), $user->getRoles())
+    ) {
+      return $validTransitions;
+    }
+
+    // Act on moderated content that belongs to a department.
```
arlina-espinoza, line 143: We aren't checking departments anymore, correct?

web/modules/custom/sfgov_utilities/src/Routing/RouteSubscriberAlter.php
```diff
@@ -0,0 +1,24 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterRoutes(RouteCollection $collection) {
+
+    // Allow the moderation module to modify access for the node edit form.
+    if ($route = $collection->get('entity.node.edit_form')) {
+      $route->setRequirement('_custom_access', 'sfgov_moderation.access_checker::access');
+    }
```
arlina-espinoza, line 21: Is this just needed to restrict access to the edit form on translations?

arlina-espinoza (approved):
I'm so glad this is moving forward. Thanks @zakiya, it looks really good. I just posted a couple of small questions, but I think it's good to merge. We should add a follow up task to remove the group module from the codebase once this is pushed all the way to prod.

web/modules/custom/sfgov_utilities/src/Routing/RouteSubscriberAlter.php
zakiya: Yes. I thought about naming it more specifically but never got around to it.

web/modules/custom/sfgov_moderation/src/StateTransitionValidation.php
zakiya: Correct. I'll update the comments. Thanks!

aekong (approved):
so glad to be rid of the group stuff!

PR #579, opened 2020-06-19
Healthcheck fixes

henryjiang-sfgov:
Fixes style defects listed on [Health Check](https://docs.google.com/spreadsheets/d/1iu-FEGRa6V0cQed-OUZ3ZvD4-MfSgsUjLW3sIk2Iuvs/edit#gid=0). Added cookies function to capture field values.

Created JS function to implement Form Render Options. The reason is to work on elements outside of the form.io form context, not possible via CSS.

Use case: Users don't want to type in all the form fields every time they visit.
User impact: User can now see their previous entered form field values and make adjustment as necessary. 

**Files changed:**

- sfgovpl.libraries.yml -- added definition for sfgov-formio 
- sfgov-formio.js -- moved from twig, added new function and refactored code
- paragraph--form-io.html.twig -- moved code to sfgov-formio.js
- _node-form-page.scss -- added css to fix the defects

PR #948, opened 2021-07-09
Delete bad form button hack

shawnbot:
This particular bit of CSS is breaking buttons on our forms, and there's no simple way to work around it:

https://github.com/SFDigitalServices/sfgov/blob/7d2a61e4688072b53bbed96959ca09efc79a452f/web/themes/custom/sfgovpl/src/sass/node/_node-form-page.scss#L53-L55

@henryjiang-sfgov I'm also curious about [this bit of CSS](https://github.com/SFDigitalServices/sfgov/blame/4fe838fbde04d37ab32ad19e8f596567270f70aa/web/themes/custom/sfgovpl/src/sass/node/_node-form-page.scss#L14-L15) that conflicts with formio-sfds's styles:

```css
.formio-sfds .flex-auto {
  flex: 1 1 auto;
}
```

Any idea what 6798c2280ef329eec7bb5ee1a48d3003caa4a5d3 (part of #579) was all about?

### DIFF
--- a/web/themes/custom/sfgovpl/src/sass/node/_node-form-page.scss
+++ b/web/themes/custom/sfgovpl/src/sass/node/_node-form-page.scss
@@ -50,15 +50,15 @@
 
   .form-page {
     @include contain-1090;
-    button:hover {
-      border-bottom: 3px solid;
-    }
   }
+
+  // FIXME: this should not be here
   .formio-sfds {
-    .flex-auto{
+    .flex-auto {
       flex: 0 1 auto;
     }
   }
+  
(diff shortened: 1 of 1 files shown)


aekong:
was that button hover part of an old requirement for the healthcheck form? i'm not familiar with any ui patterns we have where we want to add a bottom border on hover, at least not off the top of my head.  i'm not seeing this behavior on the actual form page either: https://test-sfgov.pantheonsite.io/node/1077.

propose to delete it.  :yolo:

aekong (approved):

PR #956, opened 2021-07-20
Translation fixes

aekong:


web/themes/custom/sfgovpl/includes/paragraph.inc
```diff
@@ -42,8 +42,12 @@ function sfgovpl_preprocess_paragraph(&$variables) {
       $variables['sbsId'] = $paragraph->getParentEntity()->id();
       $referencedEntities = $paragraph->get('field_transaction')->referencedEntities();
       if(count($referencedEntities) > 0) {
-          $txId = $referencedEntities[0]->id();
```
zakiya, line 45: Starting with line 41 I would move this phrase out to I would move this stanza out sfgovpl_preprocess_paragraph__process_step. Also what's the significance of the tx prefix? Some comments may also be useful here. Are you creating a link to the first referenced entity?

zakiya (approved):
Made some recommendations but no show-stoppers here.

web/themes/custom/sfgovpl/includes/paragraph.inc
aekong: the tx prefix was just to indicate that this is a transaction node.  i'll add some comments and move this chunk to a more specific hook.  thanks!

zakiya (approved):

PR #957, opened 2021-07-21
[SG-953] Add department title and redirect for each department news view page

bookworm2000:
Add department title in the news feed for each department.

Redirect the page from `news/dept_id` to `news/dept_name`.


web/modules/custom/sfgov_event_subscriber/sfgov_event_subscriber.module
```diff
@@ -0,0 +1,11 @@
...
+
+function sfgov_event_subscriber_views_post_build(&$view) {
+  // Testing to see when this passes to event subscriber.
+  $view->args[1] = 160;
+}
+
+function sfgov_event_subscriber_views_pre_view($view, $display_id, array &$args) {
+  // Testing to see when this passes to event subscriber.
+  $args[1] = 160;
+}
```
zakiya, line 11: Add new line at end of file.

arlina-espinoza (commented):
Hi @bookworm2000 ,

I have a couple of ideas of how this could be implemented without handling the redirects in custom code.

One way it to do it is using _views_ on the news page display, by replacing the department ID contextual filter with a department title filter (and configuring it so it accepts it all lowercase and replacing spaces with dashes). If so, we'd also need to create an update function that creates a bulk of redirects from current paths to the new ones.

Another option would be to use path aliases. Create an alias for every department node with the correct aliased route. This would not break any existing paths and the way the site is setup, I believe it will always redirect to the aliased path if they exist.

Let me know your thoughts or if you want to chat.

bookworm2000:
@arlina-espinoza Ah, I like the idea of using the aliases! I'll go in that direction today

bookworm2000:
@arlina-espinoza This is working locally now. Code definitely looks better not including all that event subscriber stuff.

bookworm2000:
@arlina-espinoza Not sure why this is rendering correctly for the `/news/160` page but not for the `/news/topics/794` page. Do you see anything obvious I am doing incorrectly?

web/themes/custom/sfgovpl/includes/views.inc
```diff
@@ -16,7 +17,24 @@ function sfgovpl_preprocess_views_view__meetings__page(&$variables) {
...
+  $nid = !empty($variables['view']->args[0]) ? $variables['view']->args[0]: NULL;
+  if ($nid) {
+    $node = Node::load($nid);
+    if ($node) {
+      $variables['field_dept'] = $node->getTitle();
+      $variables['banner_news'] = 'department';
+    }
+  }
+}
+
```
arlina-espinoza, line 34: Wondering if this works for other languages other than English?

web/themes/custom/sfgovpl/templates/views/views-view--news--page-1.html.twig
```diff
@@ -37,11 +37,20 @@
 ]
 %}
 
+{% if banner_news == 'department' and field_dept %}
+  {% set banner_subtitle = field_dept %}
+{% elseif banner_news == 'topic' and field_topic %}
+  {% set banner_subtitle = field_topic %}
```
arlina-espinoza, line 43: This should also be moved to the other PR for topics.

arlina-espinoza (changes requested):
Thanks @bookworm2000 ,

This is looking good, but we also need to add node create/edit/delete hooks to correspondingly sync up the news alias when a new node is created/edited, so that the work applies to future content.

As another improvement, the alias only applies to English - translating a page defaults to the node ID path (example: `/es/news/217`). I'd think we'd also need aliases for the other languages, but you'll need to test it out. :sweat_smile:

web/themes/custom/sfgovpl/templates/views/views-view--news--page-1.html.twig
bookworm2000: removed from this PR

web/themes/custom/sfgovpl/includes/views.inc
bookworm2000: Miraculously this does work, after testing on several languages.

web/modules/custom/sfgov_departments/src/SFgovDepartment.php
```diff
@@ -93,6 +95,43 @@ public static function createOrUpdate(NodeInterface $node) {
       $department_node = $node->getTranslation('en');
     }
 
+    // Add redirects for news view pages.
```
bookworm2000, line 98: @arlina-espinoza The delete works fine, but I'm not thrilled with the create and edit content. I don't know how to get this in better shape.

web/modules/custom/sfgov_departments/src/SFgovDepartment.php
```diff
@@ -93,6 +95,43 @@ public static function createOrUpdate(NodeInterface $node) {
...
+          $position = strrpos($news_alias, '/');
+          $department_alias = $position === FALSE ? $news_alias : substr($news_alias, $position + 1);
+          $news_alias = '/news/' . $department_alias;
+          if ($retrieved_alias != $news_alias) {
+            $path_alias = \Drupal\path_alias\Entity\PathAlias::create([
+              'path' => '/news/' . $node->id(),
+              'alias' => '/news/' . $department_alias,
+              'langcode' => 'en',
+            ]);
+            $path_alias->save();
```
arlina-espinoza, line 129: I think we should edit the alias object to match the latest. And we should create a redirect from the old url to the new one (the Redirect module might already do this because of the setting `Automatically create redirects when URL aliases are changed.`?)

web/modules/custom/sfgov_departments/src/SFgovDepartment.php
arlina-espinoza: @bookworm2000 Could you rebase this against the latest `develop`? I think this file was deleted as part of the work of removing the group module. If so, you can move this into a hook on another custom module.

arlina-espinoza (changes requested):
Thanks @bookworm2000 , I've added some comments. LMK if you want to discuss further. Thanks!

web/modules/custom/sfgov_departments/src/SFgovDepartment.php
bookworm2000: Rebased against develop. Yup that file was nuked so cannot use it!

bookworm2000:
@arlina-espinoza I spent a lot more time today refining this and have tested it against new/updated/deleted department and topic nodes. Hopefully this looks better!

arlina-espinoza (approved):
LGTM :)

PR #958, opened 2021-07-21
[SG-1311] Fix problem with addresses not getting proper h4 tag

bookworm2000:
For the case when a location address has a location name, but department and organization names are left blank.


zakiya (approved):

zakiya (approved):
LGTM!

PR #964, opened 2021-07-30
update /vaccine-sites microservice URL

shawnbot:
The vaccine sites microservice has [moved](https://github.com/SFDigitalServices/vaccination-site-microservice/pull/229) from Vercel to Heroku, so the URL is changing. We should get a proper CNAME (`vaccine-sites.api.sf.gov`?) and nix the `herokuapp.com` part entirely.

@zakiya I noticed that the URL is referenced in two places, so I changed it in both.


zakiya:
The url will need to be updated manually on production at https://sf.gov/admin/config/service/vaccine_microservice_settings

The active config is stored in the database for QLess and the Vaccine microservices and not the regular config files. 

cc: @aekong @arlina-espinoza @bookworm2000

zakiya (approved):

PR #967, opened 2021-08-03
[SG-1442] Fix left and right padding for resource collection intro text

arshad:
Fixes SG-1442


arlina-espinoza (approved):
Thanks @arshad

PR #971, opened 2021-08-03
[SG-1445] Fix problem with node cloning not starting from node latest revision

bookworm2000:
I created a new issue in the quick_node_clone issue queue since there were no existing ones that covered this bug:
https://www.drupal.org/project/quick_node_clone/issues/3226576

Here is the patch I wrote to fix the bug:
https://www.drupal.org/files/issues/2021-08-03/get-correct-revision-3226576-2.patch

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -248,6 +248,9 @@
             "drupal/block_field": {
                 "Error when updating block fields": "https://www.drupal.org/files/issues/2019-05-02/block_field-available-blocks-2861670-34.patch"
             },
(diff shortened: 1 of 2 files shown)


arlina-espinoza (dismissed):
LGTM :+1:

aekong:
@bookworm2000 can you resolve the lock file conflicts (prob just need to update with `develop`).  thanks!

aekong (approved):

PR #972, opened 2021-08-04
[SG-1408] Add check for resource description field

bookworm2000:
URL addresses and email addresses break the tiles, so check this value when a user adds to the form, and throw a form error if found.

### DIFF
--- a/web/modules/custom/sfgov_campaigns/sfgov_campaigns.module
+++ b/web/modules/custom/sfgov_campaigns/sfgov_campaigns.module
@@ -5,6 +5,57 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\job_scheduler\Entity\JobSchedule;
 
 /**
+ * Implements hook_form_alter().
+ */
+function sfgov_campaigns_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  if ($form_id === 'node_campaign_edit_form') {
+    $form['#validate'][] = 'sfgov_campaigns_validate_resource_description';
+  }
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function sfgov_campaigns_validate_resource_description(&$form, FormStateInterface $form_state) {
+  // Retrieve the campaign node to check.
+  $node = $form_state->getFormObject()->getEntity();
+  if (!$node->hasField('field_contents') || $node->field_contents->isEmpty()) {
+    return;
+  }
+  if ($node->getType() != 'campaign') {
+    return;
+  }
+
+  // Navigate down through the fields and related entity fields to find Resource Description.
+  $node_contents = $node->get('field_contents');
+  foreach ($node_contents as $node_content) {
+    if ($node_content->entity->hasField('field_resources')) {
+      $resources = $node_content->entity->get('field_resources');
+      foreach ($resources as $resource) {
+        if ($resource->entity->hasField('field_content')) {
+          $content_subfields = $resource->entity->get('field_content');
+          foreach ($content_subfields as $content_subfield) {
+            if ($content_subfield->entity->hasField('field_description')) {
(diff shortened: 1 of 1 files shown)


web/modules/custom/sfgov_campaigns/sfgov_campaigns.module
```diff
@@ -4,6 +4,57 @@ use Drupal\Core\Entity\EntityInterface;
...
+      foreach ($resources as $resource) {
+        if ($resource->entity->hasField('field_content')) {
+          $content_subfields = $resource->entity->get('field_content');
+          foreach ($content_subfields as $content_subfield) {
+            if ($content_subfield->entity->hasField('field_description')) {
+              $field_description_value = $content_subfield->entity->get('field_description')->value;
+              // Finally check the description field, and throw error if needed.
+              $has_http = (strpos($field_description_value, 'http://') !== FALSE) ? TRUE : FALSE;
+              $has_https = (strpos($field_description_value, 'https://') !== FALSE) ? TRUE : FALSE;
+              $regex = '/[_a-z0-9-]+(\.[_a-z0-9-]+)*@[a-z0-9-]+(\.[a-z0-9-]+)*(\.[a-z]{2,3})/';
```
arlina-espinoza, line 43: This might be too specific (what if the email ends in `.info`, for example). Validating emails is always a pain, but in this case, you just want to make sure that something that looks like an email isn't entered, so I think the following regex would work just find: `\S+@\S+`. I have not test it it, but let me know if it does work.

web/modules/custom/sfgov_campaigns/sfgov_campaigns.module
```diff
@@ -4,6 +4,57 @@ use Drupal\Core\Entity\EntityInterface;
...
+              $has_at = preg_match($regex, $field_description_value) ? TRUE : FALSE;
+              if ($has_http || $has_https || $has_at) {
+                $error_message = t('The campaign resource Resource Description field cannot contain a web address.');
+                $error_message .= t(' The field also cannot contain the @ symbol unless it has spaces on either side');
+                $form_state->setErrorByName("field content", $error_message);
+              }
```
arlina-espinoza, line 49: Let's split this up. 
```suggestion
              if ($has_http || $has_https) {
                $error_message = t('The campaign resource Resource Description field cannot contain a web address.');
                $form_state->setErrorByName("field content", $error_message);
              }
              if ($has_at) {
                $error_message .= t('The campaign resource Resource Description field cannot contain the @ symbol unless it has spaces on either side.');
                $form_state->setErrorByName("field content", $error_message);
              }
```

arlina-espinoza (changes requested):
Hi @bookworm2000 ,
I've added a couple of suggestions. Thanks!

web/modules/custom/sfgov_campaigns/sfgov_campaigns.module
bookworm2000: @arlina-espinoza I tried a bunch of email addresses using `.com`, `.net`, `.info`, etc., and this regex finds them all. It does not flag any text strings with spaces on either side of the `@` symbol. I like this for checking but if you think I should change it I will go ahead and do so.

web/modules/custom/sfgov_campaigns/sfgov_campaigns.module
bookworm2000: Split this section up so that the error messages are more specific.

web/modules/custom/sfgov_campaigns/sfgov_campaigns.module
arlina-espinoza: If works, we can keep it :)

arlina-espinoza (approved):
LGTM, thanks @bookworm2000 :)

PR #974, opened 2021-08-05
SG-1448 - translate "Part of" label for transaction

aekong:
Fixes the translation for the "Part of" label for transactions when coming from a step by step page.

_However_, if a translation in the current language does not exist for that step by step page, the "Part of" label will default to our site's default language (en)

Additionally, there is a `notranslate` class on the containing `article` parent element, so google translate will not translate that label


arlina-espinoza (changes requested):
With the changes introduced in the PR, the "Part of" title gets lost whenever a direct link to that page is shared, because there is no referrer. In addition, if a user access that page through a direct link after a cache clear, it could stay cached that way.
 If the issue is just translating the "part of" strings, Drupal's javscript API could be used for that purpose `Drupal.t()` (see https://www.drupal.org/docs/7/api/localization-api/translating-strings-in-javascript).

aekong:
Good call.  I've updated the hook to read query params instead.  Thanks!

arlina-espinoza (changes requested):
@aekong This looks good, but I somehow don't see it working correctly on Pantheon. If you visit https://pr-974-sfgov.pantheonsite.io/zh-hant/decide-if-buying-home-city-help-right-you and then https://pr-974-sfgov.pantheonsite.io/zh-hant/decide-if-buying-home-city-help-right-you?nid=625 it won't show the "part of" section. Maybe we do have to force drupal to use the `url.query_args` cache context?

web/themes/custom/sfgovpl/includes/node.inc
```diff
@@ -148,6 +150,34 @@ function sfgovpl_preprocess_node__data_story(&$variables) {
   }
 }
 
+/**
+ * Implements hook_preprocess_node().
+ * Creates the "Part of" label when coming from a step by step page
+ */
+function sfgovpl_preprocess_node__transaction(&$variables) {
+  $request = \Drupal::request();
```
arlina-espinoza, line 158: @aekong Found how to get the cache to work!: 
```suggestion
  $variables['#cache']['contexts'][] = 'url.query_args';
  $request = \Drupal::request();
```

web/themes/custom/sfgovpl/includes/node.inc
```diff
@@ -148,6 +150,34 @@ function sfgovpl_preprocess_node__data_story(&$variables) {
...
+  $fromId = $request->query->get('nid');
+
+  if (empty($fromId)) return;
```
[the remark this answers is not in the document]
aekong: 🤦 thank you!

web/themes/custom/sfgovpl/includes/node.inc
aekong: 🎉 thank you!

arlina-espinoza (approved):
LGTM :)

PR #975, opened 2021-08-06
Entity reference unpublished

aekong:
This pr adds the [Entity Reference Unpublished](https://www.drupal.org/project/entity_reference_unpublished) and the [Embed](https://www.drupal.org/project/embed) modules in order to allow users without the permission `Bypass content access control` to reference the following unpublished content:

- content types
- taxonomy terms
- media (required manual patch)

We will also need to update the reference method for every field to `Unpublished Default` or `Unpublished Media` where appropriate.

I also snuck in a script to `drush user:create` and `drush user-add-role` to create a user for each role we have to facilitate testing in our multidev environments 😅

### DIFF
--- a/config/language/zh-hant/field.field.node.article.field_tags.yml
+++ b/config/language/zh-hant/field.field.node.article.field_tags.yml
@@ -1,2 +1,2 @@
-label: '標籤 (Tags)'
+label: Tags
 description: '輸入一個以逗號分隔的列表。例如：阿姆斯特丹, 墨西哥市, "克利夫蘭, 俄亥俄州"'
--- a/web/private/scripts/drush_create_users/create_users.php
+++ b/web/private/scripts/drush_create_users/create_users.php
@@ -27,5 +27,3 @@
   }  
 
   _test_hook_slack_notification("create users: \n" . (strlen($debug) > 0 ? $debug : '  no users created'));
-
-  
--- a/web/private/scripts/drush_create_users/create_users.php
+++ b/web/private/scripts/drush_create_users/create_users.php
@@ -8,7 +8,7 @@
   ob_end_clean();
   $json = json_decode($roles);
 
-  $pw = _get_secrets(['drush_pw']);
+  $pw = _get_secrets(['drush_pw'])['drush_pw'];
   $debug = '';
 
   foreach ($json as $role => $val) {
--- a/web/private/scripts/drush_create_users/create_users.php
+++ b/web/private/scripts/drush_create_users/create_users.php
@@ -8,6 +8,7 @@
   ob_end_clean();
   $json = json_decode($roles);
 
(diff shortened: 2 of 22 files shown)


web/private/scripts/drush_create_users/create_users.php
```diff
@@ -0,0 +1,29 @@
...
+
+  foreach ($json as $role => $val) {
+    if ($role !== 'authenticated' && $role !== 'anonymous') {
+      $machineNameRole = $role; 
+      $normalizedRole = strtolower(str_replace('_', '', $role));
+      $user = 'test.' . $normalizedRole;
+      $email = $user . '@test.com';
+      $exitStatus = null;
+      $output = null;
+      exec("drush user:create ${user} --mail=\"${email}\" --password=\"password\"", $output, $exitStatus);
```
arlina-espinoza, line 21: It will be useful to create users with each role, but I'm hesitant to explicitly set the passwords, because this is a public repo and the development environments are public (they are not password protected), and that's a security risk.
Perhaps we could tackle this issue in another way? A module that could be useful for development environments could be `https://www.drupal.org/project/masquerade`, as it allows you to switch your user temporarily to another account.

arlina-espinoza (dismissed):
Marking as approved, as everything works, but read my comment about the creation of test accounts. I think we should approach that issue in a different way.

web/private/scripts/drush_create_users/create_users.php
aekong: yeah, i thought about that for a second, but dismissed it because i figured these were dev environments and these accounts wouldn't get created in production since this script only runs on the `create_cloud_development_environment` and `sync_code` patheon workflow hooks (which according to [this doc page](https://pantheon.io/docs/quicksilver#hooks) only executes in dev/multidev envs).  

But I think you're right.  Better safe than sorry! I'll use the [secrets storage method](https://pantheon.io/docs/read-environment-config#can-i-set-my-own-environment-variables) outlined by pantheon instead.

I don't really want to add another module (and I'm not sure what security warnings we'll get from a module that isn't covered by drupal's security advisory policy)

config/language/zh-hant/field.field.node.article.field_tags.yml
```diff
@@ -1,2 +1,2 @@
-label: '標籤 (Tags)'
```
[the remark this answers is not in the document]
aekong: Yeah, I'm not sure about this one.  I pulled down the live db and did a config export to preserve live changes so I wouldn't overwrite anything after I started messing around with the entity_reference_unpublished module and changing field configs to reference unpublished things

aekong:
@arlina-espinoza i'm closing this in favor of #981 (which doesn't have the unexpected config changes)

PR #976, opened 2021-08-06
Allow editor to update vaccine sites header description

aekong:
Exposes the header description text to be user-editable.

Also updates config to match what's currently on production, which is probably unnecessary now that the header text is coming from state

### DIFF
--- a/web/modules/custom/sfgov_vaccine/src/Services/VaxValues.php
+++ b/web/modules/custom/sfgov_vaccine/src/Services/VaxValues.php
@@ -59,5 +59,16 @@ class VaxValues {
   public function setAlert($value) {
     $this->state->set('vaccine_alert', $value);
   }
+
+  public function getHeaderDescription() {
+    $header_db = $this->state->get('header_description');
+    $header_db_val = $header_db['value'];
+    $header_config_val = $this->settings('template_strings.page.description');
+    return isset($header_db) ? $header_db_val : $header_config_val;
+  }
+
+  public function setHeaderDescription($value) {
+    $this->state->set('header_description', $value);
+  }
 }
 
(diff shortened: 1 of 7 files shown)


web/themes/custom/sfgovpl/templates/components/vaccine/vaccine-widget--header.twig
```diff
@@ -4,7 +4,7 @@
       {{ template_strings.page.title|t }}
     </h1>
     <p class="lead-paragraph notranslate">
-      {{ template_strings.page.description|t }}
+      {{ header_description | raw }}
```
aekong, line 7: @arlina-espinoza now this is no longer coming from the config but from drupal state, do i get translation for free? or do i have to tack on the `|t` filter? or can i even? 😅

arlina-espinoza: It is still needed :woman_shrugging:

arlina-espinoza (changes requested):
Just needs the header description to run through `t()`. Everything else looks good.

web/modules/custom/sfgov_vaccine/src/Services/VaxValues.php
```diff
@@ -59,5 +59,17 @@ public function getAlert() {
   public function setAlert($value) {
     $this->state->set('vaccine_alert', $value);
   }
+
+  public function getHeaderDescription() {
+    $header_db = $this->state->get('header_description');
+    $header_db_val = $header_db['value'];
+    $header_config_val = $this->settings('template_strings.page.description');
+    $headerDescription = isset($header_db) ? $header_db_val : $header_config_val;
+    return t($headerDescription);
```
arlina-espinoza, line 68: It would be nice to avoid calling global functions within classes. Instead, make the class use the `StringTranslationTrait`, and call `$this->t()` here.

arlina-espinoza (approved):
Looks good, thanks @aekong . I just added a suggestion as a comment.

arlina-espinoza (approved):
All good to go! :+1:

PR #980, opened 2021-08-10
[SG-1484] Add author exposed filter to content view

bookworm2000:
Allow users who can access `/admin/content` to search for authors, which is already included as a field in the rendered view.


c3gabe (approved):
Looks good

PR #981, opened 2021-08-11
Sg 1483 ref unpublished

aekong:
This pr adds the [Entity Reference Unpublished](https://www.drupal.org/project/entity_reference_unpublished) and the [Embed](https://www.drupal.org/project/embed) modules in order to allow users without the permission `Bypass content access control` to reference the following unpublished content:

- content types
- taxonomy terms
- media (required manual patch)

We will also need to update the reference method for every field to `Unpublished Default` or `Unpublished Media` where appropriate.


arlina-espinoza (approved):
LGTM, thanks!

PR #988, opened 2021-08-17
[SG-1471] Add a warning for editors about remove button not working

bookworm2000:
For the related field which is an entity reference. The other fields in the node edit view can be removed for some reason.


bookworm2000:
This hasn't been QA'ed but here is the code

aekong:
closed in favor of #1006 , per https://sfgovdt.jira.com/browse/SG-1471?focusedCommentId=68262

PR #989, opened 2021-08-18
[SG-1477] Change permissions for report content type

bookworm2000:
The publisher role should be able to create, edit, and delete own


aekong:
closing, resolved in #999

PR #993, opened 2021-08-24
SG-1416 - Patch for undefined form index error

c3gabe:
Patch to fix undefined index errors from form handler


aekong:
Thanks @c3gabe! Was this error happening outside of profile pages? Would you happen to have a way to repro this on `test` or `dev`?

PR #999, opened 2021-08-28
Sg 1477 roles permissions from prod

aekong:
and other prod configs


arlina-espinoza (dismissed):
Approving with comments:
- Changes to the data story form: are the promote and sticky properties now getting used?
- Publisher - do they really need "administer entity browsers"? That's drupal configuration, not directly related to content.

arlina-espinoza (approved):
LGTM!

PR #1000, opened 2021-08-30
Purge unused CSS utilities

shawnbot:
This PR introduces [PurgeCSS] to our CSS build process to "purge" unused selectors from our stylesheets. Here's how it works:

- Our PurgeCSS config lives in [purgecss.config.js] so that we can use the same options for both the CLI and the postcss plugin. There are two options that do all the work for us:
    1. `rejected: true` tells PurgeCSS to collect and report information about the list of rejected ("purged") CSS when it runs
    2. `content: [...]` lists paths (mostly globs) where it should look for HTML elements, attributes, and any "token" that looks like a class name. Currently I have this set to pull in:
        * `templates/**/*.{twig,html}` to match all templates in this Drupal theme
        * `src/**/*.js` to match class names that we may use to either query or add/remove in our JavaScript
        * `../../../modules/custom/**/*.{html,inc,php,theme,twig}` to match files that might contain classes in any of our custom modules
- [postcss-purgecss]() takes the options and from our config, reads all of the files in the `content` globs listed above, and "rejects" selectors that don't match anything found in the content
- A [custom postcss reporter][purge reporter] takes the rejected messages for each file, generates a `{filename}.purged.json` alongside each `{filename}.css`, and outputs some useful info to the console

So now, when you run `npm run build-css` you should see some helpful output like this:

```
🔥 purged 53 un-referenced selectors from dist/css/critical.css
   see dist/css/critical.css.purged.json for the full list
🔥 purged 1582 un-referenced selectors from dist/css/drupal.css
   see dist/css/drupal.css.purged.json for the full list
🔥 purged 26 un-referenced selectors from dist/css/editorial.css
   see dist/css/editorial.css.purged.json for the full list
🔥 purged 3093 un-referenced selectors from dist/css/utilities.css
   see dist/css/utilities.css.purged.json for the full list
```

I've added a new `utilities.css` bundle that imports `dist/css/utilities.css` from the design system and purges utilities that aren't referenced in any of our templates. That specific `@import` rule is handled by [postcss-import], which I've configured to only inline imports whose URL ends in `.css`, so that it doesn't conflict with the Sass imports.

### Testing
I think this is a combination of gut checking the rejected selector lists and QA, but in a nutshell: We need to make sure that this isn't rejecting the wrong selectors.

If there's CSS that's being purged from bundles _besides_ `utilities.css` (which is basically purge-able by design because it includes every permutation of every utility class), then we should be deleting the corresponding (S)CSS! (Fun fact: creating this PR was actually how I discovered that all of the CSS in our `twitter.css` bundle was irrelevant, which allowed it to be deleted in #998 💪 )

#### Disabling PurgeCSS
If it looks like PurgeCSS is over-aggressively purging selectors that _are_ used, we'll need to do one or both of the following:

1. Add [`purgecess ignore` comments](https://purgecss.com/safelisting.html#in-the-css-directly) to CSS that we want to prevent from being purged.
2. Add [safe lists](https://purgecss.com/safelisting.html) to our PurgeCSS config to prevent specific selectors from being purged.

[purgecss]: https://purgecss.com/
[postcss-purgecss]: https://github.com/FullHuman/purgecss/tree/master/packages/postcss-purgecss#readme
[postcss-import]: https://github.com/postcss/postcss-import
[purge reporter]: https://github.com/SFDigitalServices/sfgov/blob/1016056b0a83e50095a01edfa870d39eecfb0409/web/themes/custom/sfgovpl/lib/postcss/purgecss-reporter.js
[purgecss.config.js]: https://github.com/SFDigitalServices/sfgov/blob/1016056b0a83e50095a01edfa870d39eecfb0409/web/themes/custom/sfgovpl/purgecss.config.js


aekong:
we may want to test the admin theme too.

also, you can see some quick screenshots of where there might be some issues: http://ronswanbot.herokuapp.com/ghost-inspector/suite-results/612d26c69bed453ad10dee52

web/themes/custom/sfgovpl/postcss.config.js
```diff
@@ -1,10 +1,18 @@
+const purgeConfig = require('./purgecss.config')
+const purgeReporter = require('./lib/postcss/purgecss-reporter')
+
 module.exports = {
   syntax: 'postcss-scss',
   plugins: [
+    require('postcss-import')({
+      filter: path => path.endsWith('.css')
+    }),
```
shawnbot, line 9: This is where we inline `@import` statements, but only for paths that end in `.css` (i.e., not Sass).

web/themes/custom/sfgovpl/purgecss.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  rejected: true,
+  content: [
+    'templates/**/*.{twig,html}',
+    'src/**/*.js',
+    '../../../modules/custom/**/*.{html,inc,php,theme,twig}'
```
shawnbot, line 6: Are there other files we should be slurping up here to look for CSS selectors (class names)? Do we have JS in custom modules that toggle classes, for instance?

web/themes/custom/sfgovpl/src/sass/node/_node-dept-card.scss
```diff
@@ -20,4 +21,4 @@
     @include fs-body-short;
     color: $c-white;
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 24:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/block/_block-search.scss
```diff
@@ -214,4 +215,4 @@ header[role="banner"] {
       width: 28%;
     }
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 218:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/components/_callout.scss
```diff
@@ -16,4 +17,4 @@
     padding: 0;
     margin: 0;
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 20:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/components/_last-updated.scss
```diff
@@ -1,6 +1,7 @@
+/* purgecss start ignore */
 .last-updated {
   @include contain-1090;
   font-style: italic;
   margin-bottom: 30px;
   opacity: .6;
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 7:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/components/_person-card.scss
```diff
@@ -21,4 +22,4 @@
     width: 160px;
     text-align: left;
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 25:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/components/cards/_featured-item.scss
```diff
@@ -20,4 +21,4 @@
   &:hover {
     background: $c-purple-4;
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 24:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/components/cards/_location.scss
```diff
@@ -19,4 +20,4 @@
     color: $c-black;
     border-radius: 0 0 8px 8px;
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 23:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/eck/_event_address.scss
```diff
@@ -14,4 +15,4 @@
   a {
     display: block;
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 18:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/field/_field-related-depts.scss
```diff
@@ -1,8 +1,9 @@
...
 .sfgov-related-depts {
   color: $c-white;
   > .title {
     @include fs-title-1;
     margin-bottom: 40px;
     color: $c-white;
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 9:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/media/_media.scss
```diff
@@ -10,4 +11,4 @@
       margin: 0;
     }
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 14:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/paragraphs/_paragraph-facts.scss
```diff
@@ -95,4 +96,4 @@
       }
     }
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 99:
```suggestion
}
/* purgecss end ignore */
```

web/themes/custom/sfgovpl/src/sass/paragraphs/_paragraph-people-3elected.scss
```diff
@@ -63,4 +64,4 @@
   .sfgov-title--city-officials {
     padding-top: 0;
   }
-}
\ No newline at end of file
+}/* purgecss end ignore */
```
shawnbot, line 67:
```suggestion
}
/* purgecss end ignore */
```

aekong:
looks pretty good!

did notice one thing based on the purged json report:

```
{
  "selectors": [
    ".views--meetings .views-row article .article--teaser__title",
     // ... more .views--meetings purged classes
  ]
}
```
It looks like the safelist purge comment was already added to [web/themes/custom/sfgovpl/src/sass/view/_view-meetings.scss](https://github.com/SFDigitalServices/sfgov/pull/1000/files#diff-fbbf433ec1b2210c4360663ca5e377c2575f0561abe1ef6237ef281b3047bf43R1), so not sure why the upcoming meetings view listing is off:

| `test`  | `pr-1000` |
| ------------- | ------------- |
| Screen Shot 2021-08-31 at 12 05 45 AM | Screen Shot 2021-08-31 at 12 05 53 AM  |

I thought it might be that the twig templates set the classnames in an array and then output them in a twig expression, but that's happening everywhere else, too,so not sure why these selectors are being purged.

web/themes/custom/sfgovpl/purgecss.config.js
aekong: looks like we're toggling some classes here (for admin theme it seems): https://github.com/SFDigitalServices/sfgov/blob/purgecss/web/modules/custom/sfgov_admin/js/src/sfgov-admin.js#L30

shawnbot: Thanks, I added `../../../modules/custom/**/*.js` to the glob in e22447a2.

shawnbot: I hope we're not using those classes as style hooks! For now (as of dfb4972 I think) we're just telling PurgeCSS to ignore all of the SCSS files and only purge utilities from our custom build of `utilities.css` from the design system. So even if we are styling `.use-ajax` or something, those styles shouldn't be purged (and we can confirm that they're not from the build output).

I think we'll want to revisit what to include as "content" for this purging process when we're ready to start un-ignoring those SCSS files, but I'm open to just doing it now if you all think it's a good idea.

aekong:
i'm good for this to get merged in, thanks for putting it together! do you think we should add the bit from your pr description:

> Disabling PurgeCSS
> If it looks like PurgeCSS is over-aggressively purging selectors that are used, we'll need to do one or both of the following:
> 
> Add purgecess ignore comments to CSS that we want to prevent from being purged. You can run scripts/add-purgecss-comments on one or more files to safe-list them entirely:
> 
> scripts/add-purgecss-comments path/to/foo.scss path/to/bar.scss
> Add safe lists to our PurgeCSS config to prevent specific selectors from being purged.

to the readme? I'm hoping we'll steer clear of writing too much more custom css and hopefully start using more of the design system utilities, but while the design system matures we'll probaby have to, and it'd be good to have the heads up when building locally

aekong (approved):

PR #1004, opened 2021-09-09
SG-1381 Update all resources sections to use resource entity or node

robdecker:


web/modules/custom/sfgov_profiles/sfgov_profiles.module
```diff
@@ -46,12 +46,15 @@ function sfgov_profiles_entity_presave($entity) {
...
-    }
-    elseif ($entity->get('field_profile')->target_id != $entity->get('field_profile')->target_id) {
-      Cache::invalidateTags($profile_node->getCacheTags());
+    $targetId = $entity->get('field_profile')->target_id;
+    if (!empty($targetId)) {
+      $profile_node = \Drupal::entityTypeManager()->getStorage('node')->load($entity->get('field_profile')->target_id);
+      if ($entity->isNew()) {
+        Cache::invalidateTags($profile_node->getCacheTags());
+      }
+      elseif ($entity->get('field_profile')->target_id != $entity->get('field_profile')->target_id) {
```
arlina-espinoza, line 55: This compares an item with itself. Not sure what the purpose was originally.

arlina-espinoza (changes requested):
@aekong I noticed an inconsistency in a variable comparison. I'll request changes, but let me know if you think it works as expected and I can approve.
Also, have we tested at least a couple of times the migration?

arlina-espinoza (approved):
Looks good! :crossed_fingers:

PR #1018, opened 2021-09-24
1519: turn off update messages for non-admins

rloos289:


rloos289:
Admittedly a bit of a weird approach on this one, but it felt like the only other option was to remove the message by searching for the entire string which I feel like would be much more brittle. Definitely open to suggestions.

arlina-espinoza (commented):
I did a little more digging and it seems like the message is only getting displayed to users with the "administer site configuration" permission. From the latest message on the ticket, they also want to show it to DS role, so we might need to duplicate the original logic (https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/modules/update/update.module#L64-L126) into a custom module and disable the original hook. Might be good to create a custom permission for that too.

web/modules/custom/sfgov_user_alerts/sfgov_user_alerts.module
```diff
@@ -0,0 +1,90 @@
...
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Site\Settings;
+use Drupal\update\UpdateFetcherInterface;
+use Drupal\update\UpdateManagerInterface;
+
+
+/**
+ * @file
+ * Primary module hooks for SF Gov User Alerts module.
+ */
```
arlina-espinoza, line 16: File comment should come before `use` statements.

web/modules/custom/sfgov_user_alerts/sfgov_user_alerts.module
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+
+
+/**
+ * Implements hook_module_implements_alter().
+ */
+function sfgov_user_alerts_module_implements_alter(&$implementations, $hook) {
+  if ($hook === 'page_top') {
+    unset($implementations["update"]);
```
arlina-espinoza, line 88:
```suggestion
    unset($implementations['update']);
```

arlina-espinoza (dismissed):
Thanks @rloos289:) I'm approving but I added a couple of small code style suggestions.

PR #1043, opened 2021-10-14
create users for multidev testing

aekong:
adds a script to create users for each role and updates `pantheon.yml` to execute that script on multidev creation and code sync pantheon hooks.

this should better facilitate ci role-specific e2e testing


josh-chou (approved):
👍

PR #1050, opened 2021-10-18
Add Amplitude event logging for GitHub events

shawnbot:
This is the first step of SG-1550.

The idea here is to collect some metrics on our work. I'm particularly interested in how long PRs take to review and merge, but I'm also really interested in how long CI runs and deployment to Pantheon take, so that we can better understand where we can eke out some more speed (so we can work more effectively).

GitHub Actions workflow changes don't run on pull requests, so we will need to merge this PR to start collecting events. We may need to follow up with some tweaks to the Amplitude event data, but I think this will be a good start.

### DIFF
--- a/.github/workflows/amplitude.yml
+++ b/.github/workflows/amplitude.yml
@@ -1,19 +1,15 @@
 name: Amplitude events
 on:
-  # - check_run
-  # - check_suite
   - deployment_status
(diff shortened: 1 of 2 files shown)


aekong:
🙌

aekong (approved):

PR #1060, opened 2021-10-22
SG-1532: form translate source plugin

rloos289:


web/modules/custom/sfgov_formio/src/FormioFieldProcessor.php
```diff
@@ -0,0 +1,83 @@
...
+      // reference field. If that field exists, make a call to the referenced
+      // formio doc and set each field within as a new field in the translation.
+      if ($formio_paragraph->hasfield('field_formio_data_source')) {
+        $formio_data_source = $formio_paragraph->get('field_formio_data_source')->value;
+
+        // If the results aren't in cache, set the cache.
+        $cid = 'formio_data:nid:' . $formio_paragraph->get('parent_id')->value;
+        if (!\Drupal::cache()->get($cid)) {
+          if ($formio_data = $this->getFormioData($formio_data_source)) {
+            \Drupal::cache()->set($cid, $formio_data);
```
arlina-espinoza, line 32: This will only get cleared with a global cache clear. I think we should make the length of the cache configurable as an extra parameter in the module admin form (https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Cache%21CacheBackendInterface.php/function/CacheBackendInterface%3A%3Aset/8.2.x).
Another option is to keep this permanent in the cache, but expire this item upon paragraph save.

PR #1071, opened 2021-11-02
SG-1532  form translation source plugin

rloos289:
Note, this is both SG-1532 and SG-1534 (Getting the frontend to work)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -109,6 +109,7 @@
         "drupal/telephone_formatter": "~1.0",
         "drupal/telephone_validation": "^2.1",
         "drupal/tmgmt": "^1.12",
+        "drupal/tmgmt_google": "^1.0",
         "drupal/toc_api": "^1.0",
         "drupal/toc_filter": "^2.0",
         "drupal/twig_tweak": "^2.0",
--- a/web/modules/custom/sfgov_formio/src/Plugin/Field/FormioFieldProcessor.php
+++ b/web/modules/custom/sfgov_formio/src/Plugin/Field/FormioFieldProcessor.php
@@ -4,14 +4,13 @@ namespace Drupal\sfgov_formio\Plugin\Field;
 
 use Drupal\Core\Field\FieldItemListInterface;
 use Drupal\tmgmt_content\DefaultFieldProcessor;
-use Drupal\tmgmt_content\MetatagsFieldProcessor;
 use Drupal\Component\Utility\UrlHelper;
 use Drupal\Core\Cache\Cache;
 
 /**
- * Field processor for the text field.
+ * Field processor for the Formio Json Content text field.
  */
-class FormioFieldProcessor extends MetatagsFieldProcessor {
+class FormioFieldProcessor extends DefaultFieldProcessor {
 
   /**
    * {@inheritdoc}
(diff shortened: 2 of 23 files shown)


arlina-espinoza (dismissed):
This is great! I just added a couple of small suggestions, but I think it's really clean and neat. Thanks @rloos289 !

web/modules/custom/sfgov_formio/src/Plugin/Field/FormioFieldProcessor.php
```diff
@@ -0,0 +1,87 @@
...
+    $node_tags = $parent->getCacheTags();
+    $cid = 'formio_data:' . $node_tags[0];
+    $data = [];
+    if (\Drupal::cache()->get($cid)) {
+      $data = \Drupal::cache()->get($cid)->data;
+    }
+    elseif ($parent->get('formio_url')->value) {
+      $formio_url = $parent->get('formio_url')->value;
+      $data = (array) json_decode(file_get_contents($formio_url));
+      \Drupal::cache()->set($cid, $data, CACHE::PERMANENT, $node_tags);
```
shawnbot, line 83: It looks like this is setting the cached JSON field permanently, which is not what we want. When a form.io form changes (after people edit it on form.io), we want to be able to re-request translations in Drupal using new strings from the API.

I think we want a very _short_ TTL on these cache entries so that we don't end up with stale strings to translate. Is that possible? Or am I misinterpreting the use of `CACHE::PERMANENT` here? (If so, sorry!)

shawnbot: From Slack:
> Also, RE Cache tags: Adding the node cache tags in that line makes it so that the cache invalidates on node save 😉 

@rloos289 Ideally the cache would invalidate when translations are requested. Is that possible, or can we even just not cache the JSON at all?

rloos289: The json needs to be cached because these translation classes run way more times than you think they would and without the caching it gets really slow. This is only caching the english label and value which won't change from translation to translation.

composer.json
```diff
@@ -108,6 +108,7 @@
         "drupal/telephone_formatter": "~1.0",
         "drupal/telephone_validation": "^2.1",
         "drupal/tmgmt": "^1.12",
+        "drupal/tmgmt_google": "^1.0",
```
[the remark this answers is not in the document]
shawnbot: I suggest that we keep this for testing. What's also nice about having Google Translate available as a provider is that we can use it more explicitly as a first pass on a form and fix up bad translations in TMGMT rather than sending the whole thing off to Lionbridge.

web/modules/custom/sfgov_formio/src/Plugin/Field/FormioFieldProcessor.php
shawnbot: Thanks for walking me through this. I agree that it's necessary!

shawnbot (approved):
This works great. We're going to get the Google Translate TMGMT provider working out of the box in a separate PR, but with it enabled we can at least test this feature by installing it at `/admin/tmgmt/translators` and then submitting form page translations to it, then checking the output. 🚀

PR #1072, opened 2021-11-02
Sg 1564 mobile profile

aekong:
fixes some view issues on the small screen profile:

- Several text styles should use the equivalent mobile text style. 
  - Name should be Display 1 mobile
  - The City Employee job title should be Big description mobile
  - Additional City roles should be Title 2 mobile
  - Contact  should be Title 1 mobile
  - On External profiles, the job title should be Title 3 mobile

- The spotlight image has some display issues
- Featured items are getting more spacing than the spotlight. Should just be 20px left and right.

### DIFF
--- a/web/themes/custom/sfgovpl/package-lock.json
+++ b/web/themes/custom/sfgovpl/package-lock.json
@@ -1723,12 +1723,12 @@
       "dev": true
     },
     "axios": {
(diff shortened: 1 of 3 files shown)


arlina-espinoza (approved):
:+1:

PR #1073, opened 2021-11-03
VAX-265: Add tags to vaccination sites with kids5to11.allowed === true

shawnbot:
The goal here is to add a visible "tag" to sites that have vaccine doses for kids age 5 to 11. They should look like this:

image


config/sfgov_vaccine.settings.yml
```diff
@@ -1,4 +1,4 @@
-api_url: 'https://sfgov-vaccine-sites.herokuapp.com/api/v1/appointments'
+api_url: 'https://sfgov-vaccine-sites-pr-252.herokuapp.com/api/v1/appointments'
```
shawnbot, line 1: This **must** change back before merging!

```suggestion
api_url: 'https://sfgov-vaccine-sites.herokuapp.com/api/v1/appointments'
```

aekong (dismissed):

aekong (approved):
approving again, after api url revert

PR #1076, opened 2021-11-04
add custom db config to increase limits on log file and packet size

aekong:


rloos289:
Thank You! I was  running into exactly this

aekong:
Yeah, it'll alleviate the local lando issues we're having, but the root of it, I _think_, is lionbridge/tmgmt constantly logging messages in the db. Also, we're unable to delete old jobs because we've switched providers, so they're just sitting they're clogging up the db 😞 

I opened an issue here: https://www.drupal.org/project/tmgmt/issues/3246772

Was planning to work on a patch, haven't gotten around to it yet

arlina-espinoza (approved):
Thanks @aekong !

aekong:
follow up task to dig into tmgmt to patch deleting jobs without a provider: https://sfgovdt.jira.com/browse/SG-1572

PR #1081, opened 2021-11-05
Add notranslate classes to vaccine sites header + cards

shawnbot:
I'm adding two `notranslate` clases to the `/vaccine-sites` page so that translators can more easily identify what needs to be (human) translated:

1. In the header (top blue banner)
2. To the outermost `<div>` for each site card, and removing the `notranslate` classes on children
3. To the "Showing {count} results" text above the sites list
4. Around the alert paragraph

### DIFF
--- a/web/themes/custom/sfgovpl/templates/components/vaccine/vaccine-widget--alert.twig
+++ b/web/themes/custom/sfgovpl/templates/components/vaccine/vaccine-widget--alert.twig
@@ -1,5 +1,5 @@
 {% set text %}
-  <p>
+  <p class="notranslate">
     {{ alert|t }}
   </p>
 {% endset %}
--- a/web/themes/custom/sfgovpl/templates/components/vaccine/vaccine-widget--alert.twig
+++ b/web/themes/custom/sfgovpl/templates/components/vaccine/vaccine-widget--alert.twig
@@ -1,7 +1,7 @@
 {% set text %}
-  <p class="notranslate">
+  <div class="notranslate">
     {{ alert|t }}
-  </p>
+  </div>
 {% endset %}
 
 {% include '@theme/alert.twig' with {
--- a/web/themes/custom/sfgovpl/templates/components/vaccine/vaccine-site--booking.twig
+++ b/web/themes/custom/sfgovpl/templates/components/vaccine/vaccine-site--booking.twig
@@ -19,7 +19,6 @@
     {% set fancy_text = 'Appointments available as of'|t %}
     {% set btn_classes = [
       'sfgov-cta-button__container',
-      'notranslate',
       'vaccine-site__booking-button'
     ] %}
     {% set btn_text = 'Book an appointment'|t %}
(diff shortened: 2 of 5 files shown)


aekong (dismissed):

web/themes/custom/sfgovpl/templates/components/vaccine/vaccine-widget--alert.twig
```diff
@@ -1,5 +1,5 @@
 {% set text %}
-  <p>
+  <p class="notranslate">
```
shawnbot, line 2: @ant I was hoping that this would target the alert text, but no dice. After GT runs the markup looks like this:

```html
<div class="alert sfgov-translate-lang-es" data-exp="2021-11-05" data-style="department">
  <div class="alert--container sfgov-translate-lang-es">
    <p class="notranslate sfgov-translate-lang-es">
    </p>
    <p class="sfgov-translate-lang-es">
      <span class="sfgov-translate-lang-es"><font style="vertical-align: inherit;"><font style="vertical-align: inherit;">Los niños de 5 a 11 años pueden vacunarse, pero necesitan una cita. </font><font style="vertical-align: inherit;">Muchos sitios tienen inscripciones para mayores de 12 años.</font></font></span>
    </p>
    <p class="sfgov-translate-lang-es"></p>
  </div>
</div>
```

In other words, the `{{ alert|t }}` part on the next line looks like it's rendering empty, and the alert text is actually going into the second `<p>`, which lacks the `notranslate` class.

Would it be possible to modify the alert component so that it can get additional classes?

aekong: kind of strange how that markup is being generated.  i think it's because the input for that text allows some formatting and the wysiwyg is wrapping text with paragraph tags.  And because the template has a `<p>` tag, it becomes a `<p>` within `<p>`, so chrome is "fixing it" by separating and guessing where that class goes.

if we wrapped with a `<div class="notranslate">` instead of `<p>`, it might be ok? Otherwise, allowing the `alert` component to take an array of classes would be fine too

shawnbot: Ah, good call! I'll try that.

aekong:
```suggestion
  <div class="notranslate">
```

aekong: sorry should have made a suggestion 😬 

it works for me locally

web/themes/custom/sfgovpl/templates/components/vaccine/vaccine-widget--alert.twig
```diff
@@ -1,5 +1,5 @@
...
     {{ alert|t }}
   </p>
```
aekong:
```suggestion
  <div class="notranslate">
    {{ alert|t }}
  </div>
```

aekong (approved):

PR #1085, opened 2021-11-09
Safelist margin + padding CSS utilities

shawnbot:
In #1000 we started purging "unused" code from our generated CSS, which saves our users from downloading a bunch of code. Some forms (Shared Spaces in particular, cc: @JimBrodbeck) have already adopted the new spacing utilities, but they're not currently taken into account in the "content" we provide to PurgeCSS to detect which CSS is used, so they're being removed.

This PR adds a [PurgeCSS safe list](https://purgecss.com/safelisting.html) to our config that preserves the "basic" spacing utilities: any class beginning with `m` or `p`, followed by an optional direction (`t`, `r`, `b`, `l`, `x`, or `y`), then `-` and one or more digits. This matches the utilities I found in the Shared Spaces form:

```
mb-10
mr-20
mt-40
my-40
p-40
```

The regular expression is not "anchored" with `^` or `$`, so this will also preserve responsive variants like `lg:mx-40`.

This should be a temporary fix, and I'd like to see what it does to the size of our CSS bundle before merging so that we can adjust the regular expression to avoid matching a ton of unused classes. But in this case I'm inclined to be more permissive so that we can use the whole suite of spacing utilities in our system.

### DIFF
--- a/web/themes/custom/sfgovpl/purgecss.config.js
+++ b/web/themes/custom/sfgovpl/purgecss.config.js
@@ -8,7 +8,11 @@ module.exports = {
   safelist: {
     greedy: [
       // preserve all "basic" margin and padding utilities (for forms)
-      /\b[mp][trblxy]?-\d+/
+      /\b[mp][trblxy]?-\d+/,
+      // background color utilities
+      /bg-(black|white|slate|blue|green|red|purple|yellow|grey)/,
+      // text color utilities
+      /text-(black|white|slate|blue|green|red|purple|yellow|grey|secondary)/,
     ]
   }
 }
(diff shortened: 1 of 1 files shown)


aekong:
@shawnbot yeah, i think that makes sense

aekong:
are we able to see the size of the css bundle? are you using cssstats? looks like it adds about 4k gzipped:

pantheon dev: https://cssstats.com/stats?url=https%3A%2F%2Fdev-sfgov.pantheonsite.io
this pr: https://cssstats.com/stats?url=pr-1085-sfgov.pantheonsite.io

which seems ok

aekong (approved):

PR #1090, opened 2021-11-11
Revert "Revert "SG-1503 - Drupal 9""

aekong:
🥴 🙃


aekong:
the majority of drupal 9 prep was merged in via #1094

PR #1095, opened 2021-11-17
update php, db, and drush in pantheon yml

aekong:
as part of drupal 9 upgrade, attempt to update php, db, and drush versions on pantheon via pantheon.yml first, then do the other upgrade related things (composer, configs, deprecated code, etc).


arlina-espinoza (approved):
LGTM 🤞

PR #1102, opened 2021-11-19
SG-1572: add drush command for deleting old jobs

rloos289:
Run `drush tmgmt_clean_xtm` to delete aborted/finished xtm/xtm_test jobs. Note that this will also attempt to delete files associated with these jobs that are not being used anywhere else, see line 670 of web/modules/contrib/lionbridge_translation_provider/tmgmt_contentapi/tmgmt_contentapi.module


aekong:
thanks so much for this! ran this locally and worked great 👍

aekong (approved):

PR #1140, opened 2021-12-29
Sg 1432 gtranslate link color

aekong:


patches/gtranslate_sfgov_links.patch
```diff
@@ -34,7 +34,7 @@ index c514b74..d5bd9f3 100644
 +                        }
 +
 +                        if ($settings->get('gtranslate_' . $lang)) {
-+                        $block_content .= '<a href data-sfgov-translate="' . $gtranslate_main_lang . '|' . $lang . '" class="gtranslate-link">' . $lang_name . '</a>';
++                        $block_content .= '<a href data-sfgov-translate="' . $gtranslate_main_lang . '|' . $lang . '" class="gtranslate-link no-underline text-small text-slate hocus:underline">' . $lang_name . '</a>';
```
aekong, line 37: @shawnbot here's an interesting situation (maybe).  The header links for machine translation are generated by a contrib module, and we're actually patching the module to add our own markup.  Adding the design system classes here don't prevent them from being purged (and we probably don't want to include that directory in our purgecss config).  Instead, I [added `hocus` and `underline` to the safelist](https://github.com/SFDigitalServices/sfgov/pull/1140/files#diff-add8544d5d53262f69ed579fa20c14b4a0428f7e20bf70f2be17fa8b9140931eR17) and [updated our media captions](https://github.com/SFDigitalServices/sfgov/pull/1140/files#diff-947380b1173e7523dc2015aa48cf4e6ff141916bd038ee4b1ceaaa4451345850R51) to use `text-small`.  Kind of hacky (and maybe unnecessary) to avoid safelisting too much? Let me know what you think 🙂

shawnbot: I think we should include the patch files in our purge config rather than adding to the safe list. Ideally, there would be _no_ safe list: forms would bring the utilities they need, and only the classes we use would be safe-listed. Can we try that, and use the safe list only if it doesn't work?

aekong: yup, i'll give that a try, thanks! 👍

aekong: it works! good call!

shawnbot (approved):
Nice work! 🚀

PR #929, opened 2021-06-23
Lint JS files with eslint + eslint-plugin-sfgov

shawnbot:
This introduces eslint and [eslint-plugin-sfgov](https://github.com/SFDigitalServices/eslint-plugin-sfgov) instead of Prettier for linting our JavaScript. I apparently couldn't help making some edits to these files while I was in there, and I'm considering redoing this without the edits so that it's easier to review.

~This is blocked by https://github.com/SFDigitalServices/eslint-plugin-sfgov/pull/1. I'd like to release `eslint-plugin-sfgov@1.0.0` and upgrade to that instead of the canary release it's using right now.~

PR #1143, opened 2022-01-05
Add eslint for linting JavaScript in sfgovpl theme

shawnbot:
I started this a long time ago in #929, but I've had to rebase and force-push so many times and the Pantheon multidev is so old that opening a new PR just seems to make sense. One big improvement over the original PR is that I've made the eslint setup work with the [ESlint VS Code extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) so we can see and fix many of the linting issues using Code's UI.

The goals here are:
1. Catch common errors before they happen
2. Establish a baseline JavaScript style that's _internally consistent_
3. Move us in the direction of more modern, concise JavaScript that runs more consistently in the browsers we target

Another big difference over the original PR is that I've held off on refactoring any code and stuck to code formatting issues and actual linting errors, like duplicate `var` declarations (using scoped `const` and `let`), `==` and `!=` instead of the more strict `===` and `!==`, and so on. I've also _demoted_ a couple of our [recommended rules](https://github.com/SFDigitalServices/eslint-plugin-sfgov#pluginsfgovrecommended) from errors to warnings, including `camelCase` variable names, dangling commas, missing Promise catch and/or return statements, and so on.

Some rules are unique to this project, but we might promote to [eslint-plugin-sfgov](https://github.com/SFDigitalServices/eslint-plugin-sfgov) eventually:

- [prefer-arrow-callback](https://eslint.org/docs/rules/prefer-arrow-callback) shrinks our code by preferring arrow functions (`addEventListener(e => ...)`) over the `function (...)` syntax _unless_ the function refers to `this` in an unbound scope
- [arrow-parens](https://eslint.org/docs/rules/arrow-parens) further shrinks arrow function callbacks by removing the parentheses around callbacks with a single argument: `(e) => ...` → `e => ...`
- [object-shorthand](https://eslint.org/docs/rules/object-shorthand) makes our code more concise by moving away from `property: function property (...)` style to `property (...)`
- [no-magic-numbers](https://eslint.org/docs/rules/no-magic-numbers) warns us about [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)). Code provides a context menu for extracting the values into scoped constants. These numbers often refer to hard-coded viewport widths that we should mark for refactoring.

I'm requesting reviews from you all because I'd love to get eyes on this from everyone who writes JS in this repo and make sure we're on the same page. These are not hard and fast rules, so if you have strong opinions, please speak up!


c3gabe (dismissed):
LGTM!

aekong:
is the demotion from errors to warnings to avoid refactoring in this branch? because i kind of want those to be errors.  but also happy to promote those things to errors when we're ready to undergo the refactoring effort in a later pr

aekong (approved):

aekong (approved):

PR #1144, opened 2022-01-06
SG-1593: remove all automatic transcript fetching functionlity

rloos289:


web/modules/custom/sfgov_video/sfgov_video.install
```diff
@@ -16,50 +16,50 @@ function sfgov_video_update_8700() {
...
+  //       // Update field.
+  //       $paragraph->set('field_text', [
+  //         'value' => $transcript_plain,
+  //         'format' => 'sf_restricted_html',
+  //       ]);
+  //       $paragraph->save();
+  //     }
+  //   }
+  // }
+}
```
aekong, line 65: can we delete this file?

rloos289: Probably best to at least keep a record of all the install functions used in case someone comes in and attempts to make new ones

aekong:
thanks for cleaning this up @rloos289!  will merge after product reviews in the pr env.

aekong (approved):

PR #1151, opened 2022-01-10
bundle sfds js with theme js

aekong:
Bundle the necessary design system js with our drupal theme js

### DIFF
--- a/web/themes/custom/sfgovpl/package.json
+++ b/web/themes/custom/sfgovpl/package.json
@@ -5,7 +5,7 @@
     "browser-sync": "browser-sync start --proxy https://sfgov.lndo.site --files dist",
     "build": "run-p build-css build-js",
     "build-css": "postcss --verbose -d dist/css 'src/sass/*.css'",
-    "build-js": "babel -d dist/js 'src/js/*.js'",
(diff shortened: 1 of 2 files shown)


web/themes/custom/sfgovpl/package.json
```diff
@@ -5,7 +5,7 @@
     "browser-sync": "browser-sync start --proxy https://sfgov.lndo.site --files dist",
     "build": "run-p build-css build-js",
     "build-css": "postcss --verbose -d dist/css 'src/sass/*.css'",
-    "build-js": "babel -d dist/js 'src/js/*.js'",
+    "build-js": "babel -d dist/js 'node_modules/sfgov-design-system/dist/js/sfds.js' 'src/js/*.js'",
```
shawnbot, line 8: I think instead of processing the (already processed) design system bundle with Babel, we could do this:

```suggestion
    "build-js": "babel -d dist/js 'src/js/*.js' && cp node_modules/sfgov-design-system/dist/js/sfds.js dist/js",
```

aekong: makes sense 👍

shawnbot (approved):
💪

PR #1154, opened 2022-01-11
Fix Amplitude event triggers from GitHub Actions

shawnbot:
Let's try this again... for SG-1550.


aekong (approved):
had a similar issue when i was trying #1150

PR #1160, opened 2022-01-14
set lionbridge creds post config-import

aekong:
This pr adds a script to 

- read lionbridge credentials from a non-vcs directory
- execute `drush config-set` to update the lionbridge config with the appropriate keys and values `after` the pantheon `sync_code` workflow completes


aekong:
@shawnbot i think this needs to get merged to run properly.  the multidev environments don't pick up `pantheon.yml` changes consistently

shawnbot (approved):
LGTM! Let's test it out on develop then?

PR #1163, opened 2022-01-20
update to core 9.3.3

aekong:
Updates core to 9.3.3: https://www.drupal.org/project/drupal/releases/9.3.3

Addresses this security vulnerability: https://www.drupal.org/sa-core-2022-001.  Relevant parts as follows:

> **Vulnerability**: 
Cross Site Scripting
>
>**Description**: 
jQuery UI is a third-party library used by Drupal. This library was previously thought to be end-of-life.
>
> Late in 2021, jQuery UI announced that they would be continuing development, and released a jQuery UI 1.13.0 version. As part of this 1.13.0 update, they disclosed the following security issue that may affect Drupal 9 and 7:
>
> - CVE-2021-41183: [XSS in the of option of the .position() util](https://github.com/jquery/jquery-ui/security/advisories/GHSA-j7qv-pgf6-hvh4)
>
>It is possible that this vulnerability is exploitable with some Drupal modules. As a precaution, this Drupal security release applies the fix for the above cross-site description issue, without making any of the other changes to the jQuery version that is included in Drupal.

This pr also updates a few other dependencies

### DIFF
--- a/composer-manifest.yaml
+++ b/composer-manifest.yaml
@@ -12,13 +12,13 @@ packages:
     chi-teck/drupal-code-generator: 1.33.1
     commerceguys/addressing: v1.2.2
     composer/ca-bundle: 1.3.1
-    composer/composer: 2.1.14
+    composer/composer: 2.2.4
     composer/installers: v1.12.0
     composer/metadata-minifier: 1.0.0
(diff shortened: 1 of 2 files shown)


hshaosf (approved):
LGTM!

PR #1168, opened 2022-01-22
SG-1622  alter event and meeting date field

rloos289:
This PR encompasses the work for both SG-1622 and 1623 (adding the new date field and theming it). 
I tried my best to wrangle the date functionality into one unified place (the formatter) since at the moment it is handled in several different ways in several different places. Also, all of the if/else statements in the twig is there so that the site can support the old field and the new field until we get the migration done.


web/themes/custom/sfgovpl/templates/node/node--event--card.html.twig
```diff
@@ -15,14 +15,23 @@
...
+        <div class="date">{{ start_date_timestamp|date("j") }}</div>
+      {% endif %}
     </div>
   {% endif %}
   <div class="event-info">
     <p class="title">{{ label }}</p>
     <p class="description">
-      {{ start_date_timestamp|date("l, F j") }}<br/>
+      {% if content.field_smart_date[0]['#date'] %}
+        {{ content.field_smart_date[0]['#date'] }}
```
arlina-espinoza, line 31: This works without a `|date()` filter, correct?

web/themes/custom/sfgovpl/templates/node/node--meeting--full.html.twig
```diff
@@ -37,11 +37,17 @@
 
 {{ content.content_moderation_control }}
 
+{% if content.field_smart_date[0]['#date'] %}
+  {% set date = content.field_smart_date[0]['#date'] %}
+{% else %}
+  {% set date = node.field_start_date.date|date('F j, Y') %}
+{% endif %}
```
arlina-espinoza, line 44:
```suggestion
{% set date = content.field_smart_date[0]['#date'] ?: node.field_start_date.date|date('F j, Y') %}
```

web/themes/custom/sfgovpl/templates/node/node--view--meetings--block-past.html.twig
```diff
@@ -3,6 +3,11 @@
 {% block node %}
   <a href="{{ url }}"{{ attributes.addClass(classes)|without('id', 'role') }}>
     <p class="title">{{ node.label }}</p>
-    {{ content }}
+      {% if content.field_smart_date[0]['#date'] %}
+        {# Custom field formatter #}
```
arlina-espinoza, line 7: Could add in the template comments the reference to the module files that provide the formatter? That way themers will have an easier time figuring out where this is coming from.

arlina-espinoza (dismissed):
Thanks @rloos289, I feel like this is an elegant solution to a sticky problem. I wouldn't worry about the multiple if/else in the templates, maybe just add to-dos to cleanup after the migration. Otherwise just added some questions or comments, but marking as approved.

config/field.field.node.event.field_smart_date.yml
```diff
@@ -0,0 +1,30 @@
...
+    - tmgmt_content
+third_party_settings:
+  tmgmt_content:
+    excluded: false
+id: node.event.field_smart_date
+field_name: field_smart_date
+entity_type: node
+bundle: event
+label: Date
+description: 'Note: If you make the time range 11:59pm-12am, it won''t display on the frontend.'
```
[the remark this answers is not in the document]
rloos289: Sortof. The smart date field has time optional, but time isn't ever actually optional when using a timestamp to store the data. So the way smart date gets around that is by setting values without a time to go from 11:59-12am, and then internally it knows not to display those. I figured I'd give the warning in here just in case

web/modules/custom/sfgov_dates/sfgov_dates.module
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ */
+function sfgov_dates_node_presave($entity) {
+  // @todo, remove this once the migration to smart_date is done.
+  if ($entity->hasfield('field_smart_date') && $entity->hasfield('field_start_date')) {
+    $smart_date = $entity->get('field_smart_date');
+    $date_value = DrupalDateTime::createFromTimestamp($smart_date->value, 'UTC');
+    $entity->set('field_start_date', $date_value->format('Y-m-d\TH:i:s'));
```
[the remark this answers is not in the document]
rloos289: I did this because Drupal saves timestrings in UTC regardless of what the site's timezone is. Unfortunately all of the database entries for the field do not define a timezone, so I can't confirm that this is correct for every entry other than the few that I tested during the course of this PR. Though I feel pretty confident this is fine, the timezone for the site is LA, and especially for this one I see little reason for that to have ever changed :P

web/themes/custom/sfgovpl/templates/node/node--event--card.html.twig
rloos289: Yep. I set the config to use the `SfgovCardDate` formatter which does all of the date formatter stuff for you. A sub-goal here was removing the date filters from twig because that feels clunky to me.

aekong (approved):

PR #1170, opened 2022-01-26
Fixing top spacing on Staff Section

Daltyn06:
https://sfgovdt.jira.com/browse/SG-1604

### DIFF
--- a/web/themes/custom/sfgovpl/src/sass/node/_node-dept.scss
+++ b/web/themes/custom/sfgovpl/src/sass/node/_node-dept.scss
@@ -432,7 +432,6 @@
 
   .sfgov-dept-about-people {
     @include contain-1090;
-    padding: 80px 0;
 
     .__people {
       background: $c-blue-1;
--- a/web/themes/custom/sfgovpl/src/sass/node/_node-dept.scss
+++ b/web/themes/custom/sfgovpl/src/sass/node/_node-dept.scss
@@ -432,7 +432,7 @@
 
   .sfgov-dept-about-people {
     @include contain-1090;
-    padding-top: 80px;
+    padding: 80px 0;
 
     .__people {
       background: $c-blue-1;
--- a/web/themes/custom/sfgovpl/src/sass/node/_node-dept.scss
+++ b/web/themes/custom/sfgovpl/src/sass/node/_node-dept.scss
@@ -432,6 +432,7 @@
 
   .sfgov-dept-about-people {
     @include contain-1090;
+    padding-top: 80px;
 
     .__people {
       background: $c-blue-1;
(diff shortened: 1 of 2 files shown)


web/themes/custom/sfgovpl/src/sass/node/_node-dept.scss
```diff
@@ -432,6 +432,7 @@
 
   .sfgov-dept-about-people {
     @include contain-1090;
+    padding: 80px 0;
```
shawnbot, line 435: I think the margin utilities already add the spacing we need, so you should be able to just delete this line. Right now it's doubling up the space in between elements:

image


```suggestion
```

web/themes/custom/sfgovpl/templates/node/node--department--full.html.twig
```diff
@@ -223,7 +223,7 @@
   {% endif %}
 
   {% if content.field_people|render %}
-    <div class="sfgov-dept-about-people">
+    <div class="sfgov-dept-about-people mt-80 mb-80">
```
shawnbot, line 226: You can shorten this to `my-80`, which applies 80px of vertical margin (both top and bottom).
```suggestion
    <div class="sfgov-dept-about-people my-80">
```

web/themes/custom/sfgovpl/templates/node/node--department--full.html.twig
```diff
@@ -235,7 +235,7 @@
 
 {% if hasAddress or hasPhone or hasEmail %}
   <a name="contact"></a>
-  <div id="sfgov-dept-contact" class="sfgov-dept-section sfgov-dept-contact sfds-layout-container">
+  <div id="sfgov-dept-contact" class="sfgov-dept-contact sfds-layout-container mt-80 mb-80">
```
shawnbot, line 238: Same as above:

```suggestion
  <div id="sfgov-dept-contact" class="sfgov-dept-contact sfds-layout-container my-80">
```

web/themes/custom/sfgovpl/templates/node/node--department--full.html.twig
```diff
@@ -235,7 +235,7 @@
 
 {% if hasAddress or hasPhone or hasEmail %}
   <a name="contact"></a>
```
shawnbot, line 237: It doesn't look like it's causing problems right now, but we might want to move this anchor into the `#sfgov-dept-contact` div immediately below so that it's clearer how the margins collapse.

shawnbot (changes requested):
Thanks for this, @Daltyn06! I left some notes about the space doubling up and some tweaks that we could make. Let me know if you have any questions.

shawnbot (approved):
Thanks, looks great! 🚀

PR #1178, opened 2022-02-01
noticket: setup local phpcs and phpcbf commands

rloos289:


aekong:
it works and i love it! thanks @rloos289 !! ❤️ ❤️ ❤️

web/tests/phpcs.xml
```diff
@@ -0,0 +1,36 @@
...
+  <file>.</file>
+  <arg name="extensions" value="php,module,inc,install,test,profile,theme,css,info,txt,md,yml"/>
+  <!-- Change this value to 7 if you want to check Drupal 7 code. -->
+  <config name="drupal_core_version" value="9"/>
+
+  <!-- If you have Coder installed locally then you can reference the Drupal
+  standards with relative paths. Otherwise simply use "Drupal" and
+  "DrupalPractice. -->
+  <rule ref="Drupal"></rule>
+  <rule ref="DrupalPractice"></rule>
```
shawnbot, line 14: Correct me if I'm wrong, but it looks like [DrupalPractice](https://www.drupal.org/project/drupalpractice) is abandoned:

image

image

Can we safely delete this?

web/tests/phpcs.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ruleset name="myproject">
+  <description>PHP CodeSniffer configuration for myproject development.</description>
```
shawnbot, line 3:
```suggestion
<ruleset name="SF.gov">
  <description>PHP CodeSniffer configuration for SF.gov development.</description>
```

web/tests/phpcs.xml
```diff
@@ -0,0 +1,36 @@
...
+  "DrupalPractice. -->
+  <rule ref="Drupal"></rule>
+  <rule ref="DrupalPractice"></rule>
+
+    <!-- Disable some squiz array rules -->
+  <rule ref="Squiz.Arrays.ArrayBracketSpacing"/>
+  <rule ref="Squiz.Arrays.ArrayDeclaration">
+    <exclude name="Squiz.Arrays.ArrayDeclaration.NoKeySpecified"/>
+    <exclude name="Squiz.Arrays.ArrayDeclaration.KeySpecified"/>
+  </rule>
```
shawnbot, line 21: Do we really want to disable these, or is this just to test that we can disable individual rules?

web/tests/phpcs.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
shawnbot, line 1: Honest question: Is there a best practice that suggests putting this file in `web/tests` as opposed to the project root?

web/tests/phpcs.xml
rloos289: DrupalPractice is still a valid sniff within the coder module https://git.drupalcode.org/project/coder/-/tree/8.3.x/coder_sniffer

web/tests/phpcs.xml
rloos289: It can probably go in either safely. I usually put it in there because this also ends up being a good place for phpunit.xml or other config xml files.

web/tests/phpcs.xml
rloos289: Looks like thats been added to the core xml file in web/core/phpcs.xml.dist, so yeah we can safely remove it

web/tests/phpcs.xml
shawnbot: The one argument I would make for putting it in the project root is that I _think_ it means you can nix the `--standard` arguments from the `phpcs` and `phpcbf` commands. From [the docs](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-default-configuration-file):

> If you run PHP_CodeSniffer without specifying a coding standard, PHP_CodeSniffer will look in the current directory, and all parent directories, for a file called either `.phpcs.xml`, `phpcs.xml`, `.phpcs.xml.dist`, or `phpcs.xml.dist`. If found, configuration information will be read from this file, including the files to check, the coding standard to use, and any command line arguments to apply.

web/tests/phpcs.xml
shawnbot: Okay, works for me!

web/tests/phpcs.xml
rloos289: True, but is removing that something we actually want? Imo its a lot clearer to devs what standard is being used if we explicitly call out the path there. Especially since this is a slightly unusual pattern, normally its something like `-`--standard=Drupal,Drupalpractice`

shawnbot: I think so! The presence of `phpcs.xml` seems like a reasonable queue to me, but I'm coming from the world of JS and Node conventions where we litter our project roots with a bunch of `*.config.js` files, so... 😉 

Another plus of having it in a "standard" location is that VS Code (and probably other editor-specific) extensions that run `phpcs` and `phpcbf` will just work out of the box. I tried [this one](https://marketplace.visualstudio.com/items?itemName=ValeryanM.vscode-phpsab) and it didn't work until I added a `.vscode/settings.json` with `"phpsab.standard": "${workspacePath}/web/tests/phpcs.xml"`. Moving it to the project root allowed me to delete the `settings.json` and it just worked. ✨ 

@aekong @arlina-espinoza do you all have opinions on this? I'm happy to be wrong here!

aekong: i don't really have a preference, though i'm in the same boat as @shawnbot and usually expect tooling config to be in the project root

rloos289: Alright, y'all have convinced me. Moved

phpcs.xml
```diff
@@ -0,0 +1,29 @@
...
+  <rule ref="DrupalPractice"></rule>
+
+  <!-- Exclude markdown and txt files. -->
+  <exclude-pattern>*md</exclude-pattern>
+  <exclude-pattern>*txt</exclude-pattern>
+  <!-- Exclude contrib. -->
+  <exclude-pattern>../modules/contrib</exclude-pattern>
+  <!-- Exclude auto-generated theme files. -->
+  <exclude-pattern>../themes/custom/sfgovpl/node_modules</exclude-pattern>
+  <exclude-pattern>../themes/custom/sfgovpl/dist</exclude-pattern>
```
shawnbot, line 23: A couple of things here: 1) fix the relative path to `modules/contrib`; 2) exclude all `node_modules` and `dist` directories, since we may have those in custom modules as well.

```suggestion
  <exclude-pattern>web/modules/contrib</exclude-pattern>
  <!-- Exclude npm-managed directories -->
  <exclude-pattern>**/node_modules</exclude-pattern>
  <!-- Exclude generated asset directories -->
  <exclude-pattern>**/dist</exclude-pattern>
```

phpcs.xml
```diff
@@ -0,0 +1,29 @@
...
+
+  <!-- If you have Coder installed locally then you can reference the Drupal
+  standards with relative paths. Otherwise simply use "Drupal" and
+  "DrupalPractice. -->
+  <rule ref="Drupal"></rule>
+  <rule ref="DrupalPractice"></rule>
+
+  <!-- Exclude markdown and txt files. -->
+  <exclude-pattern>*md</exclude-pattern>
+  <exclude-pattern>*txt</exclude-pattern>
```
shawnbot, line 18: We could simplify this by leaving `txt` and `md` out of the extensions list:
```suggestion
  <arg name="extensions" value="php,module,inc,install,test,profile,theme,css,info,txt"/>
  <!-- Change this value to 7 if you want to check Drupal 7 code. -->
  <config name="drupal_core_version" value="9"/>

  <!-- If you have Coder installed locally then you can reference the Drupal
  standards with relative paths. Otherwise simply use "Drupal" and
  "DrupalPractice. -->
  <rule ref="Drupal"></rule>
  <rule ref="DrupalPractice"></rule>
```

PR #1179, opened 2022-02-02
1656: setup lando phpcs and phpcbf

rloos289:
@aekong @shawnbot. Closing the original PR, this one includes everything we discussed and has a cleaner commit history + ticket number reference


shawnbot (approved):
Thanks, this is great! 🚀

PR #1207, opened 2022-03-01
Amplitude page view json

aekong:
- patch amplitude to fix the missing annotation error which was preventing changes from being saved
- update the yaml config for page view event `properties` key to a string


shawnbot (approved):
Ugh, I'm sorry. The newlines in the `properties` field are my fault! Thanks for fixing this.

aekong:
no sorry, i probably reviewed and LGTM'd it, on me too :high-five:

PR #1239, opened 2022-03-24
SG-1079: fixing range cost display to show inline.

Daltyn06:


web/themes/custom/sfgovpl/src/sass/node/_node-event.scss
```diff
@@ -243,5 +243,11 @@
     }
 
   }
+
+  .paragraph--type--cost {
+    .cost-range {
+      display: inline-flex;
+    }
+  }
```
aekong, line 251:
```suggestion
```
can we use the design system utility class `inline-flex` on the markup instead? pls see suggestion below

PR #1245, opened 2022-03-31
SG-1663 - Translated forms should not have translated URLs

aekong:
Update pathauto pattern to use the token `[node:source:title]` for form pages


shawnbot (approved):
I tested this by adding a Spanish translation of [this form](https://pr-1245-sfgov.pantheonsite.io/es/submit-your-accessible-business-entrance-forms/form), and it works! 🎉

PR #1246, opened 2022-03-31
update purge finished jobs to 30 days

aekong:


aekong:
change already in production.  pr is to preserve change on next deploy

PR #1251, opened 2022-04-05
Translation fixes

aekong:
Fixes:
- https://sfgovdt.jira.com/browse/SG-1683
- https://sfgovdt.jira.com/browse/SG-1695
- https://sfgovdt.jira.com/browse/SG-1700


web/private/scripts/translation-providers/credentials.php
```diff
@@ -0,0 +1,51 @@
...
+
+$providers = [
+    $lionbridge,
+    $google,
+];
+
+foreach($providers as $provider) {
+    $status = '';
+    $output = [];
+    $creds = file_get_contents(
```
josh-chou, line 36: $creds could be false if there was an failure in file_get_contents (ie. file doesn't exist).
Might want to check != false and handle accordingly before running exec.

josh-chou (approved):
minor suggestion

josh-chou (approved):

web/private/scripts/translation-providers/credentials.php
aekong: thanks!

PR #1255, opened 2022-04-07
SG-1709 - Meeting and event URLs does not include date (for recurring instances)

aekong:
Updates the pathauto url alias patterns for the `Event` and `Meeting` content type to use the new date field


aekong:
Configuration updated in production.  This pr persists the change in config files for next release

PR #1256, opened 2022-04-11
SG-1650 - disable old department about field

aekong:


aekong:
this change is already in production.  this pr serves to persist the change in config files

PR #1272, opened 2022-05-05
(hopefully) make it easier to log in with different roles

aekong:
* adds an accordion on the login page for logging in as the different roles that exist on sf.gov
* enabled only for multidev environments, not enabled for `live` or `content`

image


aekong:
updated the details and list elements to use design system utility classes

PR #1282, opened 2022-05-16
create users for non-live environments (including test)  on deploy

aekong:
trying to create test users in the pantheon `test` environment, but the `test` environment pantheon deploy hooks are specific to `test` _and_ `live`, there isn't one specific for `test` only.  

this pr modifies the existing script to create test users to only execute in non-live pantheon environments, so that we can auto-execute the script with the pantheon `deploy` workflow hook


josh-chou (approved):
lgtm

PR #1304, opened 2022-06-01
[SG-1749] - Fix reading order accessibility issues on Profiles

paboden:
Changed profile name for persons to use H1 instead of div

Instructions:
1. Clear cache
2. Review person node (full page) and confirm name is in H1


c3gabe (approved):
LGTM

PR #1305, opened 2022-06-01
[SG-1747] - Fix reading order accessibility issues on "Document" section on Resource Collections.

paboden:
Section headings on Resource collections need new heading tags to have more accessibility compliance:

“Documents” title should be H2 (instead of H3)
“Document Subsection Title” should be H3 (instead of H4)
“Resources” title should be H2 (instead of H3)
“Resources Subsection Title” should be H3 (instead of H4)
“Data” title should be H2 (instead of H3)
“Data Story Subsection” title should be H3 (instead of H4)

Instructions:
1. Clear cache
2. Review a resource collection(s) with the different elements above and inspect their tags


paboden:
Added a fix to the color of the h1 in the banner, and a fix to a potential php error if the documents in that section don't return a viable file object.

aekong (approved):

PR #1312, opened 2022-06-02
SG-1724 follow up - consider attached file and document url

aekong:


paboden (dismissed):
Looks good. Approved.

paboden (approved):
Url generation looks good

PR #1313, opened 2022-06-03
Sg-1740  form translation v2

rloos289:
Changes include:

- Checking for valid json on “Formio Render Options”
- Checking for valid URL and valid json on “Data Source”
- Remove the add item/remove buttons on “additional form string” for non-english
- Restructuring internal logic to make validation and migration easier (this means that its still worth testing the general workflow)


arlina-espinoza (dismissed):
Thanks @rloos289, this is looking good in a service :) I've just added some extra validations to avoid possible PHP notices, but the structure of the service and calls all looks good.

aekong (approved):

PR #1325, opened 2022-06-15
Update /vaccine-sites with new tags for pediatric dosages

shawnbot:
See [Figma](https://www.figma.com/file/ZpEVJd6jGnRDdf5SMrypDQ/Vaccines?node-id=2034%3A7822) for reference design. See VAX-273 for AC.

- [x] Remove walk-up and drive-through access mode options (preserving wheelchair accessibility)
- [x] Add "Vaccines for kids" drop-down with age ranges: `6 months to 4 years`, `5 to 11 years`, and `12 to 17 years` ("none" option label: `Age range`)
- [x] Add brand-specific dosage tags to site cards, e.g. `Moderna 5 and under`, `Pfizer 4 and under`, etc. (get `dosages` from SFDigitalServices/vaccination-site-microservice#284 with age ranges overlapping the selected "Vaccines for kids" range and add tags for those with brand names)
- [x] Activate the "Vaccines for kids" filter with JS
- [x] Refactor a ton of JS and CSS ✨ 
- [x] Merge #1338 with automatically importing translations
- [x] [Test it thoroughly!](https://pr-1325-sfgov.pantheonsite.io/vaccine-sites)
- [x] Revert the API URL change on [this line](https://github.com/SFDigitalServices/sfgov/blob/e818926e1501ea465e5f8ebdc373709ea6c6d485/config/sfgov_vaccine.settings.yml#L1)

We've decided to address the remaining issues with the address autocomplete and distance search in another pass.


.circleci/config.yml
```diff
@@ -1,8 +1,7 @@
 # https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs
 defaults: &defaults
   docker:
-    # TODO: use "build-tools-ci:6.x" when we're on Node 16.x
-    - image: quay.io/pantheon-public/build-tools-ci@sha256:ea1a5b8486b80b431bf82861e0402a3f4bd37dc0e913da16f2a35c89e88857b5
+    - image: quay.io/pantheon-public/build-tools-ci@sha256:2cebe0db5c0cf5682ad14fa28550435a402425d2c9ef3c000e343db62ad93fde
```
shawnbot, line 4: This puts us on Node 16, which helps with npm workspaces.

web/modules/custom/sfgov_vaccine/js/filter_sites.js
```diff
@@ -1,329 +1,319 @@
...
+        }
+
+        const maxDistance = parseFloat(radiusInput.val())
+        const origin = {
+          lat: parseFloat(locationField.attr('data-lat')),
+          lng: parseFloat(locationField.attr('data-lng'))
+        }
+        const showDistance = userLocation && [maxDistance, origin.lat, origin.lng].every(n => !isNaN(n))
+        if (showDistance) {
+          filters.push(site => site.distance.value <= maxDistance)
```
shawnbot: Look ma, no regular expressions! The way that filtering works now is that we conditionally build up an array of filtering functions depending on which filters are "active", then in the jQuery `.filter()` callback we test _every_ filter function against the site data (now parsed from `data-site`).

web/modules/custom/sfgov_vaccine/js/filter_sites.js
```diff
@@ -1,329 +1,319 @@
...
+      const radiusInput = $('[name=radius]')
+
+      const allSites = $('.vaccine-site')
+        .each(function () {
+          const site = getSiteData(this)
+          site.distance = {
+            origin: NO_ORIGIN,
+            value: undefined
+          }
+          jQuery.data(this, 'site', site)
```
shawnbot, line 32: Probably better to be consistent here:

```suggestion
          $(this).data('site', site)
```

web/modules/custom/sfgov_vaccine/js/filter_sites.js
```diff
@@ -1,329 +1,304 @@
...
-          wheelchair_chkBox.datatest = "1";
-        } else {
-          wheelchair_chkBox.datatest = "";
+        const ageRangeString = pedAgeSelect.val()
+        if (ageRangeString) {
+          const ageRange = ageRangeString.split('-').map(str => parseFloat(str))
+          if (ageRange.every(n => !isNaN(n))) {
+            filters.push(site => site.dosages.some(dosage => rangesOverlap(ageRange, dosage.ages)))
+          } else {
+            // console.warn('Bad age range:', ageRange)
```
jacksongonzales, line 95: we should prob either uncomment or remove the else clause

web/modules/custom/sfgov_vaccine/src/Controller/VaccineController.php
```diff
@@ -247,16 +222,36 @@ private function getSiteLanguageText($access_data) {
   private function getSiteEligibilities($site_data) {
 
     $printed = [];
+
+    if (isset($site_data['dosages']) && is_array($site_data['dosages'])) {
+      foreach ($site_data['dosages'] as $dosage) {
+        if ($dosage['ages'][1] <= 5) {
+          $brand = $this->t($dosage['brand']);
+          // Format < 0 decimal numbers (really just 0.5) with underscores
```
jacksongonzales, line 230: typo: this should say "< 1" 🙂

shawnbot:
```suggestion
          // Format < 1 decimal numbers (really just 0.5) with underscores
```

web/modules/custom/sfgov_vaccine/src/Form/FilterSitesForm.php
```diff
@@ -71,39 +71,50 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
     ];
 
-    // Single checkboxes - kids 5 to 11
-    $form['container']['toggle']['items']['single_checkboxes']['kids5to11'] = [
-      '#type' => 'checkbox',
-      '#title' => $this->t($this->settings('form_strings.kids5to11')),
-      '#default_value' => FALSE,
-    ];
+    $pediatric_options = [];
+    $option_values = $this->settings('form_strings.pediatric.options');
```
jacksongonzales, line 91:
```suggestion
    $option_values = $this->t($this->settings('form_strings.pediatric.options'));
```

might have this syntax wrong 😬

jacksongonzales: ah i see this should be happening on line 93 below...however I'm still seeing this

shawnbot: Yeah, we don't have translations for all of those strings yet, and I don't think there's a way to "pre-populate" them via the configs. It looks like we can import the strings from the spreadsheet with [drush local:import](https://www.drush.org/latest/commands/locale_import/):

```sh
terminus drush locale:import --type=customized es path/to/strings.es.po
# repeat for zh and fil language codes
```

Now I just need to work out how to generate PO files from [the spreadsheet](https://docs.google.com/spreadsheets/d/1sYtN_z294Nr5dZHgX-7EnmcI6z9U6oYH9fNDjHehM98/edit#gid=0) 😉

web/modules/custom/sfgov_vaccine/js/filter_sites.js
shawnbot: I'm on the fence about this. One the one hand, I think we should allow ourselves to have console messages in dev and remove them in production, but _only_ if we're able to log them some other way. On the other hand, having a minifier just strip out the comment and the empty `else` block would give us way more flexibility and not have to solve the production logging challenge.

Either way, I vote we punt on this 😉

web/modules/custom/sfgov_vaccine/src/Form/FilterSitesForm.php
shawnbot: Figured it out in #1338 🎉

web/modules/custom/sfgov_vaccine/js/filter_sites.js
```diff
@@ -1,329 +1,304 @@
...
+      }
+
+      function rangesOverlap ([x1, x2], [y1, y2]) {
+        return x1 <= y2 && y1 <= x2
+      }
+
+      function inCityBounds ({ lat, lng }) {
+        return (
+          lng < -122.93 || lng > -121.54 ||
+          lat < 37.0000 || lat > 38.0200
```
aekong, line 299:
```suggestion
          (lng > -122.93 || lng < -121.54) &&
          (lat > 37.0000 || lat < 38.0200)
```
previous comparison was checking for out of city bounds.  since we're now comparing in city bounds, think the comparison operators have to flip

shawnbot: Nice catch! I'll try that.

web/modules/custom/sfgov_vaccine/js/filter_sites.js
```diff
@@ -1,329 +1,304 @@
...
+      submitButton.on('click', event => {
+        event.preventDefault()
+        leftColumn.fadeOut(0)
+
         // Clear GPS data, check for autocomplete entry
-        if(!$("#edit-location")[0].getAttribute("data-place-changed")) {
-		  // Remove previous GPS, ***BUG*** not working for invalid address handling
-          document.getElementById("edit-location").removeAttribute("data-lat");
-          document.getElementById("edit-location").removeAttribute("data-lng");
+        if (!editLocation.attr('data-place-changed')) {
```
jacksongonzales, line 43: not sure if this is helpful, but i'm noticing that the `data-place-changed` attr is hanging around when entering invalid addresses (along with the `data-lat/-lang` ones)

shawnbot: Yeah, there's some super funky stuff going on with this. (Hence the **BUG** comment above, I guess? 😂 )

jacksongonzales (approved):
lgtm!

PR #1334, opened 2022-06-17
[SG-1746] - Link to archive sfgov.org from department home pages

paboden:
Added 2 new fields to departments:
- Archive URL (field_archive_url)
- Archive date (field_archive_date)

Setup output next to the Request public records output (at the bottom of the content).

Instructions:
1. Clear cache
2. Import configuration
3. Confirm that the 2 new fields are added to departments by going to the Department content type
4. Edit an existing department or create a new one
5. Set the archived url and date values
6. Confirm they display as desired.

### DIFF
--- a/config/core.entity_form_display.node.department.default.yml
+++ b/config/core.entity_form_display.node.department.default.yml
@@ -9,6 +9,8 @@ dependencies:
     - field.field.node.department.field_address
     - field.field.node.department.field_alert_expiration_date
     - field.field.node.department.field_alert_text
+    - field.field.node.department.field_archive_date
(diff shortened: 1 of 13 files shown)


paboden:
It seems to be failing because the test can't create a department. Yet I can log in to the PR site and create one. I think the new required setting on "Request public record" checkbox is throwing a false positive in the test

aekong:
> It seems to be failing because the test can't create a department. Yet I can log in to the PR site and create one. I think the new required setting on "Request public record" checkbox is throwing a false positive in the test

yup, we can fix that text post-merge to `develop`

aekong (approved):

PR #1337, opened 2022-06-20
[SG-1007] - "See more events" is linking to the wrong list view

paboden:
Rebuild of all events pages (upcoming and past), and added new page displays for topic and department filtered events (both upcoming and past).

Required reworking of the view templates and preprocess code to account for the new displays.

Reworked the "See more events" link on topic and department pages to automatically go to the filtered event upcoming pages. Filtered past event pages can be clicked to on the upcoming page.

Instructions:
1. clear cache.
2. make sure config is synced for the "Events" view.
3. Visit /events and confirm that upcoming and past events still display as they should.
4. Visit a topic page and a department page, that have at least 4 upcoming events (create some if needed) and click on the "See more events" link
5. Should take you to a url that follows this pattern - /events/topic/(title-of-topic) or /events/department/(title-of-department)
6. Can manually go to topic pages like "/events/topic/immigrants", "/events/topic/taxes", or department versions like "/events/department/treasure-island-development-authority"

Remember, event content may need to be created to be able to see upcoming events because most topics and departments do not have many/any upcoming events.

Notes: It is possible that if a topic or department title contains some abnormal characters (like & or $), it might not process cleanly for the contextual filtering.
Some known fixes are in place, but stuff is possible, however unlikely it may be.


paboden:
Note: I set the wrong jira ID in the first commit message, so one of the links is off. It should link here - https://sfgovdt.jira.com/browse/SG-1007

paboden:
Pantheon has hit a limit for environments so this will have to wait

aekong (approved):

PR #1342, opened 2022-06-25
Translation and /vaccine-sites docs, clean up, Dangerfile

shawnbot:
This is an attempt to document #1080 and #1338 after the rushed release. See [config/translations/README.md](https://github.com/SFDigitalServices/sfgov/blob/chore/vaccine-sites-cleanup/config/translations/README.md#readme) for the gory details.

This also introduces [Danger](https://danger.systems/js/) to enforce naming conventions for our translation files. It generates warnings if any of the following rules are broken in the `config/translations` directory:

1. If there is a `.pot` (English template), there should be a corresponding `.po` file for each language ending in `.es.po`, `.zh-hant.po`, and `.fil.po`
2. If there is no corresponding `.pot` file for a group of similarly named `.{lang}.po` files
3. If there is a `.po` file without a language code in the filename (`.{lang}.po`)
4. If the `{lang}` part of a `.{lang}.po` isn't one of the language codes we support: `es`, `zh-hant`, or `fil`

For good measure, I've also included some warnings for pull request descriptions that are either too short (less than 50 chars) and/or don't include a Jira ticket reference.

Warnings are made in a PR comment and look like this:

<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Missing translations for `config/translations/vaccine-sites.pot`: `config/translations/vaccine-sites.es.po` (Spanish), `config/translations/vaccine-sites.zh-hant.po` (Chinese), and `config/translations/vaccine-sites.fil.po` (Filipino)
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  Missing translations for `config/translations/test.pot`: `config/translations/test.es.po` (Spanish), `config/translations/test.zh-hant.po` (Chinese), and `config/translations/test.fil.po` (Filipino)
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  Missing English template `config/translations/whatever.pot` for `whatever.*.po`
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>Please include at least one Jira ticket, e.g. SG-123 or DESSYS-123.</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

[...]

PR #1346, opened 2022-06-27
CI: add compressed asset size report

shawnbot:
This adds the [compressed-size-action](https://github.com/preactjs/compressed-size-action) to the checks workflow (FKA "danger") added in #1342. [This comment](https://github.com/SFDigitalServices/sfgov/pull/1346#issuecomment-1167993790) demonstrates what it will look like until we add, modify, or delete any of our existing JS or CSS bundles.

We can either merge this into that branch (the current base, for ease or reviewing) or merge that PR first then set the base of this PR to `develop`.


.github/workflows/checks.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          node-version: 16
+      - run: npm install
+      - run: npx danger ci
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+      - uses: preactjs/compressed-size-action@v2
+        with:
+          pattern: '**/dist/**/*.{js,css}'
+          exclude: '**/node_modules/**'
```
shawnbot, line 23: Currently this looks for all `.js` and `.css` files in any `dist` directory (and excludes anything under any `node_modules`), but we might need to be more specific about this later. Note that we're not doing a `composer install` on GitHub Actions because it's not necessary to build the assets, but if we do we'll also need to exclude the composer-installed dependencies in `/vendor`.

PR #1354, opened 2022-07-05
JS linting updates

shawnbot:
For SG-1796 (and SG-1655), this PR standardizes our ESLint + Babel setup across custom modules and theme(s), following on from the monorepo setup in #1325 and cleanup in #1342. The top-level `babel.config.js` and `.eslintrc.js` control Babel and ESLint, respectively, and corresponding configs in each package (`web/{modules,themes}/custom/*` directories containing a `package.json`) just extend the root-level configs.

For this PR I've only added a new `package.json` to the `sfgov_vaccine` and `sfgov_search` modules; we'll need to add the following files to each custom module directory to adopt the standardized tooling:

- `package.json` to declare dependencies and the `build` and `lint-js` scripts
- `babel.config.js` to compile browser JS (e.g. from `src/js` to `dist/js`, which seems like a good convention)
- `.eslintrc.js` to adopt the top-level ESLint config
- `.gitignore` to exclude generated files in `dist/` from git

As of this PR, any package directory with a `lint-js` npm script **opts itself into** the linting workflow in GitHub Actions (which runs `npm run -ws --if-present lint-js`), which will fail a build if there are ESLint errors! These run in Actions so that they don't delay builds or prevent deployments, but can still raise red flags.

I've also added a check to our Dangerfile to warn us if there are dependency version mismatches between any of the packages and the root.

If you're curious, [here's a report](https://gist.github.com/shawnbot/159acabdf12bf2b4cafceecf0ce9d331) of all the violations in the `web/modules/custom/sfgov_*` JavaScript 😁


arlina-espinoza (dismissed):
Hi @shawnbot, nice to get this linting in! I think we should make the linting rules aligned with Drupal's JS coding standards (https://www.drupal.org/docs/develop/standards/javascript/javascript-coding-standards). These are the rules that I see need to be fixed: 
- All statements (except for, function, if, switch, try, while) MUST be followed by a semi-colon (`;`).
- Control structures MUST always use curly braces, even in situations where they are optional. 
- The `else` of an if/else control structure should go in it's [own line](https://www.drupal.org/docs/develop/standards/javascript/javascript-coding-standards#structures). Same for the `catch` in try/catch.
- Also, why is "'use strict'" getting removed?

arlina-espinoza (commented):
Hi @shawnbot ,

On the comment on enforcing some standards, I am all up for it, and I absolutely think that any standard we agree with is better than the free style mess we have currently 😓 . Having said that, I think the only thing I have strong feelings about, not just being a code style preference, is:
- Control structures MUST always use curly braces, even in situations where they are optional: Helps developers avoid logic errors while coding, making the blocks explicit.

It would also be good to add documentation to the main project's README on how this works, and how to get this set up working locally for development - maybe share the VS Code configuration for auto-format on save (I'm sure we can also find instructions on setting up PHPStorm as well).

Thanks!

aekong:
ran this locally, everything works well.  i like that gh actions will show us where the violations are in the files changed tab.  thanks for making this happen 🎉

aekong (dismissed):

aekong:
@shawnbot thanks for finding that! i've updated the ghost inspector test.  there's a merge conflict i resolved in #1392, just checking on that in a separate env.  if that looks good, we can that into this pr, then merge this in to develop.  thanks for all the hard work on improving our code quality!

aekong (approved):
🚀 🎉

PR #1359, opened 2022-07-08
SG-1694 - disable field_transaction from form display

aekong:


aekong:
already done in production, config exported here for next production deploy

PR #1380, opened 2022-08-17
SG-1796 fix search

aekong:
SG-1796

* attempt to fix search due to js linting updates


web/modules/custom/sfgov_search/js/sfgov_search.js
```diff
@@ -484,6 +484,8 @@
   }
 
   const search311 = new Search311()
+  window.search311 = search311
```
aekong, line 487: @shawnbot there's probably a better way to expose this object for the jsonp callback.
 https://github.com/SFDigitalServices/sfgov/blob/3b0e10aafc0aca3d04f43566b97ed2b748583279/web/modules/custom/sfgov_search/js/sfgov_search.js#L22

options include the global `drupalSettings.sfgovSearch` object which currently holds some (i believe) read-only funnelback configurations: 
```
collection: "sfgov-meta-prod"
keyword: "birth"
qie: "0.0"
```
so we _could_ just assign the search311 object to this object and update the callback to be `drupalSettings.sfgovSearch.search311` (as opposed to attaching to the `window` object, which i believe, generally, should not be done)

shawnbot: Let's try it out! Ideally we wouldn't use JSONP, but I understand how that would probably be more complicated to figure out.

aekong: ok, all set.  working ok now at https://pr-1380-sfgov.pantheonsite.io/search?keyword=birth.  pls review and merge if it looks good and we can get #1354 merged into `develop`

shawnbot (approved):
🚀

PR #1414, opened 2022-10-07
SG-1866 -- See more news link not showing on some department news listings

paboden:
SG-1866 -- See more news link not showing on some department news listings
Reordering of view displays fixed this.


paboden:
Just need to config import the view and make sure the departments block display comes after the page display

PR #1429, opened 2022-10-20
setup sfgovpl theme to use tailwind

aekong:
* add tailwind config and set up to use design system tailwind preset
* add `order` to `corePlugins`
* add `odd` to `backgroundColor` variant

### DIFF
--- a/web/themes/custom/sfgovpl/package.json
+++ b/web/themes/custom/sfgovpl/package.json
@@ -18,7 +18,9 @@
     "watch-js": "npm run build-js -- --watch"
   },
   "dependencies": {
-    "sfgov-design-system": "^2.4.0"
+    "sfgov-design-system": "^2.5.1-rc.206cb7e",
+    "tailwindcss": "^3.2.0",
+    "tailwindcss-interaction-variants": "^5.0.0"
   },
   "devDependencies": {
     "@csstools/postcss-sass": "^5.0.1",
--- a/web/themes/custom/sfgovpl/tailwind.config.js
+++ b/web/themes/custom/sfgovpl/tailwind.config.js
@@ -1,0 +1,18 @@
+/** @type {import('tailwindcss/types').Config} */
+
+const { content } = require('./purgecss.config')
+
+module.exports = {
+  presets: [
+    require('sfgov-design-system/tailwind.preset')
+  ],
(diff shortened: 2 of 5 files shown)


aekong:
@shawnbot there's some huge savings on the file size which made me wonder if we were missing utilities.  but the pr screenshots look fine.  the diff between what `develop` generates vs this pr says that lots of margin utils have been removed (like `m-12`, `m-16`, along with a bunch of others), but after a quick search in the templates, it doesn't look like those utils are being used.  i could use a pairing session a sanity check 😅

web/themes/custom/sfgovpl/package.json
```diff
@@ -18,7 +18,9 @@
     "watch-js": "npm run build-js -- --watch"
   },
   "dependencies": {
-    "sfgov-design-system": "^2.4.0"
+    "sfgov-design-system": "^2.5.1-rc.206cb7e",
+    "tailwindcss": "^3.2.0",
+    "tailwindcss-interaction-variants": "^5.0.0"
```
shawnbot, line 23: You also shouldn't need this dependency 🤞

web/themes/custom/sfgovpl/tailwind.config.js
```diff
@@ -0,0 +1,18 @@
...
+  ],
+  corePlugins: {
+    order: true
+  },
+  variants: {
+    extend: {
+      backgroundColor: ['odd']
+    }
+  },
+  content
```
shawnbot, line 17: Two things here:

1. The `@type` annotation should go right before the `module.exports` assignment to maintain that association
2. I think the utilities CSS bundle is smaller because Tailwind is removing them before they get to the "outer" purge, which has a more permissive safelist. To validate this, I think we would bring back the `purge.options` and pass in the same config from `purge.config.js`.

```suggestion
const { content, ...purgeOptions } = require('./purgecss.config')

/** @type {import('tailwindcss/types').Config} */
module.exports = {
  presets: [
    require('sfgov-design-system/tailwind.preset')
  ],
  corePlugins: {
    order: true
  },
  variants: {
    extend: {
      backgroundColor: ['odd']
    }
  },
  content,
  purge: {
    options: purgeOptions
  }
```

PR #1430, opened 2022-10-20
use tailwind safelist

aekong:
we're trying to update our current theme build to include its own tailwind config to take the design system's preset and extend it in #1429 

tailwind's internal purge is removing all the safelisted patterns before our own purgeCSS does its thing.  tailwind used to have a `purge: false` option, [but it sounds like the `purge` option for tailwind 3 has been changed to `content` and only takes an array of glob patterns for tailwind's internal purge](https://tailwindcss.com/docs/upgrade-guide#configure-content-sources) (not 100% on this 😅 ).

[there is a `safelist` option](https://purgecss.com/safelisting.html), though, so this pr takes the purgeCss safelist and drops it in the tailwind config to honor that safelist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4925,9 +4925,9 @@
       "link": true
     },
     "node_modules/sfgov-design-system": {
-      "version": "2.5.1-rc.206cb7e",
-      "resolved": "https://registry.npmjs.org/sfgov-design-system/-/sfgov-design-system-2.5.1-rc.206cb7e.tgz",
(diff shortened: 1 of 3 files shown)


aekong:
@shawnbot this gives us a more 1-1 `dist/utilities.css` with the build in our `develop` branch but it bums me out to see that `+353%` in that filesize change report 😭 .  

it looks like we safelisted some utilities for forms in #1085.  is that still needed?

regarding [the comment in that pr](https://github.com/SFDigitalServices/sfgov/pull/1085#issuecomment-964630679) about color utilities living in drupal taxonomy, i think we might actually be ok there because the taxonomy terms expect a hard-coded hex value.  sorry to dredge up old pr's ☹️

PR #1434, opened 2022-10-28
SG-1886 - fix the drop down fields help text is resolving to a blue color 

aekong:
SG-1886

* configure `sfgov_admin` module to use current build system workflow and tools
  * add to ci build script
* remove gulp
* fix hover color on editor backend autocomplete reference lists


aekong:
@shawnbot i updated the `sfgov_admin` module to use our current build workflow.  just wanted to verify that it looks close to correct.  the affected parts of the site from this module's css and js function correctly 😅

PR #1435, opened 2022-11-04
SG# 1872 --  Allow notes to be output with addresses in sidebars

paboden:
SG# 1872
Setup new view mode for address with notes, apply view mode to "in_person_location" address output template.

Config import
Clear caches
Review locations added to sidebars


paboden:
SG# 1872

paboden:
@arlina-espinoza NOTE: I set up an additional view mode for the address eck output because I wasn't sure if the address only view mode was being used outside of the in_person_location paragraph type, and outside of the sidebar area output.
The view mode is address_with_notes.

aekong (approved):

PR #1458, opened 2022-12-07
Enable form.io submission editing on form pages

shawnbot:
For PUBX-313, this update to our form.io template to editing previously submitted form.io submissions. Here's how it works:

1. When we create a form in form.io, we add a button that triggers saving the submission as a draft
1. We receive a webhook with the submission id (`_id`, specifically) and send a draft confirmation email with a "magic link"  to the form's URL appended with `?submission={{ _id }}` in the email template
1. If the form page finds a `submission` query string parameter, it appends `/submission/{submission}` to the form URL, which tells form.io to load that submission for editing

You can see how this works in [the original CodePen](https://codepen.io/shawnbot/pen/zYayWNa) and @fwextensions's [CodePen](https://codepen.io/fwextensions/pen/XWxeNNq?submission_id=64532608441c66b7bca0c731&form_id=oocEquityIncubator).

### Future enhancements
* We've discussed preventing submission edits if they're not marked explicitly as a "draft", but can't do this until we agree on what (data-wise) distinguishes a draft submission from a completed one.
* We've decided not to add support for the `page` query string parameter in this PR, but may do so in a future pass.

### DIFF
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -197,6 +197,14 @@
 
     time = Date.now()
     var dataSource = el.getAttribute('data-source')
+    var params = new URL(location).searchParams
+    if (params.submission) {
+      dataSource += `/submission/${params.submission}`
+      // options.saveDraft = true
+      Formio.setUser({
+        _id: '123'
+      })
+    }
     Formio.createForm(el, dataSource, options)
       .then(function(form) {
         amp('form.load', { load_time: Date.now() - time })
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -198,12 +198,8 @@
     time = Date.now()
     var dataSource = el.getAttribute('data-source')
     var params = new URL(location).searchParams
-    if (params.submission) {
-      dataSource += `/submission/${params.submission}`
-      // options.saveDraft = true
-      Formio.setUser({
-        _id: '123'
-      })
+    if (params.submission_id) {
+      dataSource += `/submission/${params.submission_id}`
     }
     Formio.createForm(el, dataSource, options)
       .then(function(form) {
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -197,9 +197,11 @@
 
     time = Date.now()
     var dataSource = el.getAttribute('data-source')
-    var params = new URL(location).searchParams
-    if (params.submission_id) {
-      dataSource += `/submission/${params.submission_id}`
+    let url = new URL(location)
+    var params = new URLSearchParams(url.search)
+    let submission_id = params.get('submission_id')
+    if (submission_id) {
+      dataSource += `/submission/${submission_id}`
     }
     Formio.createForm(el, dataSource, options)
       .then(function(form) {
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -197,11 +197,10 @@
 
     time = Date.now()
     var dataSource = el.getAttribute('data-source')
-    let url = new URL(location)
-    var params = new URLSearchParams(url.search)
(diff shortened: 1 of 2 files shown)


web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -197,6 +197,14 @@
 
     time = Date.now()
     var dataSource = el.getAttribute('data-source')
+    var params = new URL(location).searchParams
+    if (params.submission) {
+      dataSource += `/submission/${params.submission}`
```
shawnbot, line 202:
```suggestion
    if (params.submission_id) {
      dataSource += `/submission/${params.submission_id}`
```

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -197,6 +197,14 @@
...
+      // options.saveDraft = true
+      Formio.setUser({
+        _id: '123'
+      })
```
shawnbot, line 206:
```suggestion
```

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -197,6 +197,14 @@
...
+    }
     Formio.createForm(el, dataSource, options)
       .then(function(form) {
         amp('form.load', { load_time: Date.now() - time })
```
shawnbot, line 210: This is where we would need to abort rendering if there's a submission ID and that submission is marked as a draft.

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -197,6 +197,12 @@
 
     time = Date.now()
     var dataSource = el.getAttribute('data-source')
+    let url = new URL(location)
+    var params = new URLSearchParams(url.search)
+    let submission_id = params.get('submission_id')
+    if (submission_id) {
+      dataSource += `/submission/${submission_id}`
+    }
```
shawnbot, line 205: @Daltyn06 I noticed the fix from `params.submission_id` to `params.get('submission_id')`; thank you! The `new URLSearchParams(url.search)` bit is unnecessary, though, thanks to [URL.prototype.searchParams](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams). I'm also switching back to `var` for consistency within this file:

```suggestion
    var params = new URL(location).searchParams
    var submissionId = params.get('submission_id')
    if (submissionId) {
      dataSource += `/submission/${submissionId}`
    }
```

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -197,6 +197,11 @@
 
     time = Date.now()
     var dataSource = el.getAttribute('data-source')
+    var params = new URL(location).searchParams
+    var submissionId = params.get('submission_id')
+    if (submissionId) {
+      dataSource += `/submission/${submissionId}`
```
fwextensions, line 203: What if `dataSource` ends in a `/`?  Will Formio just ignore the extra slash?

fwextensions:
Just wanted to check where we were with this PR.  Can we get it merged?

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
shawnbot: I actually don't think the data source URLs work if there's a trailing slash. Either way, we don't include them. :smile:

web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
```diff
@@ -197,6 +197,11 @@
 
     time = Date.now()
     var dataSource = el.getAttribute('data-source')
+    var params = new URL(location).searchParams
+    var submissionId = params.get('submission_id')
```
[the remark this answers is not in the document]
shawnbot: I'm fine with that.
```suggestion
    var submissionId = params.get('submission')
```

fwextensions (approved):

fwextensions:
I'd previously tested with this URL, which used to load the submission:

https://pr-1458-sfgov.pantheonsite.io/apply-be-cannabis-equity-incubator/form?submission_id=64768bf362600e80b0750198

And now it doesn't.  But this URL with the `submission` param now does:

https://pr-1458-sfgov.pantheonsite.io/apply-be-cannabis-equity-incubator/form?submission=64768bf362600e80b0750198

So that all looks good.

PR #1475, opened 2023-01-10
SG-1940 - vaccine sites filter

aekong:
SG-1940

* remove `p-0` utility class on filter container
* remove async loading on js scripts
  * the google maps api calback parameter was reporting that `locationAutocomplete` was not found

### DIFF
--- a/web/modules/custom/sfgov_vaccine/js/location_autocomplete.js
+++ b/web/modules/custom/sfgov_vaccine/js/location_autocomplete.js
@@ -33,14 +33,14 @@ function locationAutocomplete () {
 function locationSubmit () {
   if (typeof google === 'undefined') {
     console.warn('google maps is not defined! hiding the location search UI')
-    $('#edit-distance-from').hide()
+    document.querySelector('#edit-distance-from').classList.add('hidden')
     return Promise.resolve({
       status: 'ok',
       msg: 'Google Maps is not loaded'
     })
   }
   else {
-    $('#edit-distance-from').show()
+    document.querySelector('#edit-distance-from').classList.remove('hidden')
   }
 
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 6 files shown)


web/modules/custom/sfgov_vaccine/js/location_autocomplete.js
```diff
@@ -33,14 +33,14 @@ function locationAutocomplete () {
 function locationSubmit () {
   if (typeof google === 'undefined') {
     console.warn('google maps is not defined! hiding the location search UI')
-    $('#edit-distance-from').hide()
+    document.querySelector('#edit-distance-from').classList.add('hidden')
```
shawnbot, line 36: We might want to be more error-tolerant here with [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), which will short-circuit if the query returns null:

```suggestion
    document.querySelector('#edit-distance-from')?.classList.add('hidden')
```

web/modules/custom/sfgov_vaccine/js/location_autocomplete.js
```diff
@@ -33,14 +33,14 @@ function locationAutocomplete () {
...
     return Promise.resolve({
       status: 'ok',
       msg: 'Google Maps is not loaded'
     })
   }
   else {
-    $('#edit-distance-from').show()
+    document.querySelector('#edit-distance-from').classList.remove('hidden')
```
shawnbot, line 43: Same as above:

```suggestion
    document.querySelector('#edit-distance-from')?.classList.remove('hidden')
```

shawnbot (changes requested):
Nice catch on the async script! I made a couple of suggestions to keep the DOM JS as resilient as jQuery. Otherwise this looks great. Thank you!

aekong:
thanks for the suggestions!

shawnbot (approved):
Reviewed synchronously! We're disabling the compressed-size-action report for now (in this PR and others) because it's bailing with an error about an out-of-date `package-lock.json`, and we can't replicate that locally.

PR #1561, opened 2023-05-31
chore(ci): disable lighthouse action

shawnbot:
This action has been failing for a while and I haven't had a chance to look into it. Until then, let's disable it so we don't have to see ugly red x's on all of our PRs.


aekong:
hi @shawnbot , i think it was resolved in #1537 ? the pr's here need to be updated with `develop`, should hopefully fix the failed builds on the older prs.  newer pr's seem ok

PR #1562, opened 2023-06-01
fix cost value for minimum and up

aekong:


web/themes/custom/sfgovpl/templates/paragraphs/paragraph--cost.html.twig
```diff
@@ -63,7 +63,7 @@
       {% endif %}
       {% if cost_type == 'minimum' and content.field_cost_minimum|render %}
         {% block costmin %}
-          {{ '$@min and up.'|t({'@min': content.field_cost_minimum}) }}
+          {{ '$@min and up.'|t({'@min': paragraph.field_cost_minimum.0.value}) }}
         {% endblock costmin %}
       {% endif %}
       {% if cost_type == 'range' and content.field_cost_minimum|render and content.field_cost_maximum|render %}
```
fwextensions, line 69: There are two other uses of `content.field_cost_minimum` here.  Why aren't those a problem?

aekong: the other two uses are in conditionals to check that a value exists (which may not be a good check because it just passes it to the `render` filter, which _can_ output markup, which would make the conditional evaluate to true).

i think the typical way of rendering the values in the template would be to use `content.field_cost_minimum` which is a render array, which the templating engine would output as markup along with the field value.  but because we're also wanting to translate that `$@min and up` part, we're passing that string to `t` and passing `content.field_cost_minimum` to it, which is an array, when `t` expects a string.

PR #1570, opened 2023-06-06
EPR-760 Update Formio js package

hshaosf:
We might not need to load the full javascript package from form.io. Proposing to swap `formio.full.min.js` to `formio.form.min.js`

### DIFF
--- a/web/modules/custom/sfgov_formio/README.md
+++ b/web/modules/custom/sfgov_formio/README.md
@@ -28,7 +28,7 @@ It's possible to test different versions, without having to edit the above setti
 
 | Parameter name | Example Query String | Resulting Source |
 | :--- | :--- | :---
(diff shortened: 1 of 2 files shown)


fwextensions:
As I mentioned in the other PR, the form.min file isn't actually any smaller.  This change needs more investigation.

hshaosf:
Not sure if there is a better way to cross-link PRs between repos. So just dropping a link to the related [`formio-sfds` PR](https://github.com/SFDigitalServices/formio-sfds/pull/256) here for now. :)

shawnbot (commented):
This looks good to me, assuming tests pass over in https://github.com/SFDigitalServices/formio-sfds/pull/256 and there's some manual testing of the multidev environment.

If the `form` bundle isn't any smaller than the `full` bundle, though, I think we'll need to investigate other ways to do this. For instance, we _could_ bundle only the parts of the formiojs that we need in formio-sfds. 🤔

PR #1573, opened 2023-06-29
Prevent the form page from showing a confirmation when a draft is saved

fwextensions:
This is currently being worked around with some code in the OOC forms that manipulates the event listeners array directly, but it would be better for this code to do the right thing.

### DIFF
--- a/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
+++ b/web/themes/custom/sfgovpl/templates/paragraphs/paragraph--form-io.html.twig
@@ -259,7 +259,10 @@
             }
           }
 
-          if (confirmationURL) {
+          // we want to navigate to the confirmation page only on final submission.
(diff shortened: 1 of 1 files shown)


fwextensions:
@shawnbot is there a way of testing a form with this environment?  I tried https://pr-1573-sfgov.pantheonsite.io/apply-be-cannabis-equity-incubator/form but it 404s, while this old PR does work: https://pr-1458-sfgov.pantheonsite.io/apply-be-cannabis-equity-incubator/form

This form isn't live yet, but I'm not really sure why it's in the old PR but not this one.  I guess I assumed all "not live" forms would be available in the PR environment, but I guess not.  Maybe there's some additional setup that has to happen?  Just trying to test this with a real form.

fwextensions:
nvm.  I guess you need to login to each PR environment for the non-live forms to show up.  Tested this and it looks good.

PR #1633, opened 2023-12-14
fix: fix missing formEventHandlers

aekong:


hshaosf (approved):
👍

fwextensions:
Are we sure this will fix the redirect, and not just remove the error?  All of the form event handlers were removed in this PR: https://github.com/SFDigitalServices/sfgov/commit/c956eb9c1b95d2213cfe236dde518408efe960f7#diff-d5bc080583607c0222ee1f05139afed9692a1a185a19f4c2d0152e5405732df3L115

Is the `submitDone()` handler not required?  Or `nextPage()`/`prevPage()`?

fwextensions:
And if it truly does fix it, seems like it would be clearer to just remove the `Object.keys(formEventHandlers).forEach()` loop, since it has no handlers to create.

aekong:
i believe the intent of the form handlers were for logging amplitude event data.  so you're probably right about removing the loop.  will do that now as a follow-up and push up to test in the pr env

fwextensions:
>  the intent of the form handlers were for logging amplitude event data

Some of them clearly were, but these look like they may have provided actual form functionality.  I don't know though, as I haven't really delved into how the sf.gov page handles the form events.

```js
      submitDone: function (submission) {
        return {
          submission_id: submission._id
        }
      },
      prevPage: function (data) {
        return {
          page_index: data.page,
          page_title: getPageTitle(data, this)
        }
      },
      nextPage: function (data) {
        return {
          page_index: data.page,
          page_title: getPageTitle(data, this)
        }
      },
      wizardNavigationClicked: function (page) {
        return {
          page_title: page.title
        }
      },
      wizardPageSelected: function (page, index) {
        return {
          page_title: page.title,
          page_index: index
        }
      },
      editGridAddRow: function (data) {
        return {
          component: data.component.key
        }
      }
```

aekong:
yeah, it looks like they were all added in this pr for logging amplitude events: https://github.com/SFDigitalServices/sfgov/pull/736

fwextensions:
Ah, I see.  Nicely detailed PR from @shawnbot back in the day!

PR #1689, opened 2024-06-18
fix: accessCheck for sfgov_utilities

aekong:
hotfix from prod because of cron error for deprecated method of querying


henryjiang-sfgov (approved):
approved

PR #1698, opened 2024-07-10
feature: form analytics

shawnbot:
For EPR-1294, this updates the JavaScript on our form pages to send Google Analytics / Tag Manager events (honestly, I still  don't quite understand the distinction). To make this less painful, I've pulled the form initialization out of the page template and into a separate JS file. It's not very legible, but [here's the diff](https://gist.github.com/shawnbot/b49f937e57bffa049b4332ad2a0dbd41) between the JS portion of the template and the new file.

### `measure()`
The `measure()` function, which is local to the form initialization code, supports a couple of different call signatures:

1. `measure('foo')` just sends an event named `form.foo` (the first argument is a string, which gets a `form.` prefix)
2. `measure('foo', { some: 'value' })` sends the event with additional variables (specific to that event)
3. `measure({ other: 'value' })` sets one or more variables that will [persist](https://developers.google.com/tag-platform/tag-manager/datalayer#persist_data_layer_variables) for subsequent `measure()` calls

All of these just append an object to the GA [data layer](https://developers.google.com/tag-platform/tag-manager/datalayer) via `window.dataLayer.push(...)`.

### Variables
The following variables are set on all form measurement events:

* `form.url` is set to the URL of the form schema
* `form.language` is set to the language code of the form, used for translations

### Measurement events
We should start seeing the following events in GTM:

| Event | Trigger | Variables |
| :--- | :--- | :--- |
| <pre>form.create</pre> | Before form is created | |
| <pre>form.load</pre> | After form is loaded | |
| <pre>form.nextPage</pre> | User navigates to the next form page | `page`: the index of the page |
| <pre>form.prevPage</pre> | User navigates to the previous form page | `page`: the index of the page |
| <pre>form.fileUploadStart</pre> | File upload starts | |
| <pre>form.fileUploadEnd</pre> | File upload finishes | |
| <pre>form.submit</pre> | User submits the form | `state`: the submission state; `submission.saved`: boolean indicating if the submission was previously saved |
| <pre>form.submitDone</pre> | Submission finishes successfully | |
| <pre>form.redirect</pre> | Form redirects to another page | `url`: where it's redirecting; `reason`: `confirmation` or `sfoptions.redirects`
| <pre>form.error</pre> | Form fails to initialize | `message`: the error message; `stack`: the stack trace |

### Other changes
- I fixed up the theme's babel config so that it inherits the settings from the repo root, which gives us [babel-preset-env](https://babeljs.io/docs/babel-preset-env) magic ✨ 
- I updated the babel config in the theme to remove comments from the generated files
- I tweaked our `.browserslistrc` to expand our browser compatibility coverage from [84.9%](https://browsersl.ist/#q=last+2+versions%0A%3E+1%25%0Aie+11%0A) to [91.5%](https://browsersl.ist/#q=last+2+versions%0A%3E+0.2%25%0Aie+11%0A) globally

Changes to the root `package-lock.json` are the result of running `npm audit fix`.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -17,7 +17,10 @@ jobs:
       - run: npm install
 
       - name: lint JavaScript
-        run: npm run lint-js
+        run:
+          echo "::remove-matcher owner=eslint-compact::"
+          npm run lint-js
+
 
       - name: Dangerfile
         run: npx danger ci
--- a/package-lock.json
+++ b/package-lock.json
@@ -1985,9 +1985,9 @@
       "dev": true
     },
     "node_modules/@types/cors": {
-      "version": "2.8.13",
-      "resolved": "https://registry.npmjs.org/@types/cors/-/cors-2.8.13.tgz",
-      "integrity": "sha512-RG8AStHlUiV5ysZQKq97copd2UmVYw3/pRMLefISZ3S1hK104Cwm7iLQ3fTKx+lsUH2CE8FlLaYeEA2LSeqYUA==",
+      "version": "2.8.17",
+      "resolved": "https://registry.npmjs.org/@types/cors/-/cors-2.8.17.tgz",
+      "integrity": "sha512-8CGDvrBj1zgo2qE+oS3pOCyYNqCPryMWY2bGfwA0dcfopWGgxs+78df0Rs3rc9THP4JkOhLsAa+15VdpAqkcUA==",
       "dev": true,
       "dependencies": {
         "@types/node": "*"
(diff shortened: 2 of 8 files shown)


package-lock.json
shawnbot, line 1: The diff here is the result of running `npm audit fix` in `web/themes/custom/sfgovpl`.

.browserslistrc
shawnbot, line 1: I moved this file to the repo root. It used to look like this:

```
last 2 versions
> 1%
ie 11
```

web/themes/custom/sfgovpl/package-lock.json
shawnbot, line 1: There shouldn't be a `package-lock.json` in the workspace directory.

web/modules/custom/sfgov_admin/package-lock.json
shawnbot, line 1: This file was messing with attempts to run the npm build scripts in this module's workspace.

package-lock.json
shawnbot, line 1: I ran `npm audit fix` to generate this diff

web/themes/custom/sfgovpl/src/js/formio-form-page.js
```diff
@@ -0,0 +1,237 @@
...
+          let lang = form.language
+          lang = formLanguageMap[lang] || lang
+          const actualUrl = lang && lang !== 'en'
+            ? confirmationURL.replace('{lang}', lang)
+            : confirmationURL.replace('{lang}/', '')
+          measure('redirect', {
+            reason: 'confirmation',
+            url: actualUrl
+          })
+          window.location = actualUrl
```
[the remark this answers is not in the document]
shawnbot: I dismissed this warning because the URL comes from the `data-confirmation-url` attribute, which is output in the template. To change the value of this, an attacker would need to be able to either craft a malicious form page in the Drupal admin or inject JS to modify the DOM before this script executes. In either case, the attacker would basically be able to do whatever they want, so this is not really an issue.

web/themes/custom/sfgovpl/src/js/formio-form-page.js
```diff
@@ -0,0 +1,238 @@
+/* eslint brace-style: ['error', '1tbs'] */
```
shawnbot, line 1: I opted to change the brace style (`1tbs` = [one true brace style](https://en.wikipedia.org/wiki/Indentation_style#One_True_Brace)) in this file rather than all of them.

web/themes/custom/sfgovpl/src/js/formio-form-page.js
```diff
@@ -0,0 +1,238 @@
...
+        // XXX: this condition never hits
+        // else if (field) {
+        //   // set cookie for first time
+        //   if (fieldtype === 'radio' || fieldtype === 'checkbox') {
+        //     setCookie(item, field.checked)
+        //   }
+        //   else {
+        //     setCookie(item, field.value)
+        //   }
+        // }
```
shawnbot, line 221: I commented this out because it's dead code: the `if` condition is the same as the `else if`, so the latter will never execute. I have a hunch that most of the "sfoptions" stuff can just go away, since:

1. None of the forms that use the `sfoptions` options are published; and
2. All of the ones that do use it (and which are unpublished) are related to the old Check Your Health forms, which we stopped using a very long time ago

web/themes/custom/sfgovpl/.browserslistrc
```diff
@@ -1,3 +1,3 @@
 last 2 versions
-> 1%
+> 0.2%
```
shawnbot, line 2: Going down to 0.2% increases our browser (support) coverage pretty substantially: from [84.9%](https://browsersl.ist/#q=last+2+versions%0A%3E+1%25%0Aie+11%0A) to [91.5%](https://browsersl.ist/#q=last+2+versions%0A%3E+0.2%25%0Aie+11%0A) globally.

web/themes/custom/sfgovpl/babel.config.js
```diff
@@ -1,2 +1,4 @@
 module.exports = {
+  extends: '../../../../babel.config.js',
```
shawnbot, line 2: I'm surprised we (I) didn't catch this before, but because we weren't explicitly extending the repo config we weren't getting the [babel-preset-env](https://babeljs.io/docs/babel-preset-env) transforms applied to any of our JS source files in the theme.

I took a quick look and didn't see anything in our existing JS that looks like it would trip up older browsers, but the JS that I moved out of the formio paragraph template did have at least one tagged template literal that might have been mucking things up for some (practically ancient) browsers. The upshot is that that we can now write modern JS in the theme and it'll be transformed browser-friendly ES5 automatically.

web/themes/custom/sfgovpl/sfgovpl.libraries.yml
```diff
@@ -161,3 +161,10 @@ sfgov-data-story:
     - core/jquery
     - core/drupal
     - core/jquery.once
+
+formio-form-page:
+  js:
+    dist/js/formio-form-page.js:
+      attributes:
+        # this needs to be deferred to execute after formiojs
+        defer: true
```
shawnbot, line 170: Previously we were delaying the form initialization by wrapping it in a window `load` event listener. Now that the script is loaded in a separate file, the better way to do this is with [defer](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#defer).

web/themes/custom/sfgovpl/src/js/formio-form-page.js
shawnbot, line 1: It's not very legible, but [here's the diff](https://gist.github.com/shawnbot/b49f937e57bffa049b4332ad2a0dbd41) between the JS portion of the template and this new file.

web/themes/custom/sfgovpl/.browserslistrc
jackieyoungsfgov: this is great! i'm curious what considerations went into choosing 0.2%--like why not a higher or lower number?

shawnbot: [browsersl.ist](https://browsersl.ist/) did this for me, at least in part: When I input the old config it showed a warning:

image

So I clicked "Fix" and it lowered the threshold to `0.3%`, and the coverage number went up. I tried `0.2%` and the coverage went up even further. I guess we might as well lower it to `0.1%` and get [93.8%](https://browsersl.ist/#q=last+2+versions%0A%3E+0.1%25%0Aie+11%0A)? 🤷

shawnbot: This might even be better, for [95.7%](https://browsersl.ist/#q=defaults%0A%3E+0.1%25+in+US%0A%3E+0.1%25+in+AS%0A): 

```
defaults
> 0.1% in US
> 0.1% in AS # Asia
```

Or we could just go for broke and use `> 0%`, which because of sampling yields 98.4% 😂

jackieyoungsfgov: LOL i was thinking that too but wasn't sure if there were some performance reasons not to do it 😆

.github/workflows/checks.yml
```diff
@@ -17,7 +17,10 @@ jobs:
       - run: npm install
 
       - name: lint JavaScript
-        run: npm run lint-js
+        run:
+          echo "::remove-matcher owner=eslint-compact::"
```
shawnbot, line 21: The check annotations on unchanged files were driving me nuts, so I added this [per the docs](https://github.com/actions/toolkit/blob/master/docs/commands.md#problem-matchers) to disable the eslint "problem matcher" altogether.

aekong (approved):
🎉 !!

PR #1702, opened 2024-07-12
feature: better form load errors, form file upload events

shawnbot:
This is a follow-up for #1698 that improves a few things related to event measurement:

1. When forms fail to load, the error message is more verbose and, hopefully, more useful in user reports. This at least partially addresses EPR-1243.
2. Measurements for file upload events now includes more useful variables, including file type and size

It also adds a babel plugin to strip console calls (with the exception of `warn()` and `error()`) from the built JS, and updates the paths of the JS files in the `sfgovpl.libraries.yml` to the generated files in `dist/` instead of the source files.

### Form load errors
We've seen in user reports (both in the feedback form messages and in screenshots sent to us by one user) that forms which fail to load don't produce _any_ output on the page.

My understanding of the problem is that `Formio.request()` rejects with the _text_ of failed form schema request, and the text of the response is either empty (when requests are blocked at the network level) or, as we've seen with forms that have been renamed, the hilariously un-helpful "Invalid alias".

The proposed fix is to hook into formio.js's response handling and throw a more verbose error, which forces `Formio.request()` down a different code path which includes the URL along with the HTTP response status and text on the page. The format of the message is wonky (`Could not connect to API server ({status} {statusText}): {url}`), but it should help us collect more useful information when people send us reports of forms failing to load.

### File upload measurement events
| Event | Trigger | Variables |
| :--- | :--- | :--- |
| <pre>form_file_upload_start</pre> | User navigates to the previous form page | `file_provider` (`azure`), `file_size` (size in bytes), `file_type` (mime type), and `file_name` |
| <pre>form_file_upload_end</pre> | File upload succeeds | Same as above |
| <pre>form_file_upload_error</pre> | File upload fails | Same as above, plus the error message in `error` |

All of the other form-related events have been renamed to match Google's suggested conventions:

- The event name prefix is now `form_` instead of `form.` (which nested the values under `form` in the data layer)
- Similarly namespaced data variables like `form.language` and `error.message` now use underscores instead of `.`

### Fetch plugin
Both of these are done using a [formiojs fetch plugin](https://help.form.io/developers/fetch-plugin-api), which adds a catch-all response handler and "wraps" all three of formio.js's distinct request types:

1. "Static" requests that use the global `Formio` object, which includes form schema requests
2. Form requests made by forms, namely submissions
3. File requests, which are not named in the docs but I found [in the source](https://github.com/formio/formio.js/blob/4.20.x/src/Formio.js#L618)

The wrapped request functions receive the request promise and "args" (data about the request), measures a start event, then hooks into `promise.then()` with callbacks to measure both the end and error events.


web/themes/custom/sfgovpl/src/js/formio-form-page.js
```diff
@@ -1,140 +1,252 @@
+/* eslint-disable no-console */
 /* eslint brace-style: ['error', '1tbs'] */
 ;(function () { // eslint-disable-line no-extra-semi
+  const { Formio, gtag } = window
+  const MEASURE_PARAMS = {}
+  const USE_GA = false
```
shawnbot, line 6: If we can figure out how to get `gtag()` calls to send data (maybe with a new/different Drupal plugin?) then we can just flip this to `true` and it _should_ work.

aekong (approved):
🎉

PR #1738, opened 2025-01-08
fix: redirect search to google

aekong:


shawnbot (approved):
Looks great, thank you!

